PR #7, opened 2016-10-24
[BAHIR-72][bahir-flink] support netty: pushed tcp/http connector

shijinkui:
When source stream get start, listen a provided tcp port, receive stream data from user data source.
This netty tcp source is keepping alive and end-to-end, that is from business system to flink worker directly.
1.  source run as a netty tcp and http server
2.  user provide a tcp port, if the port is in used, increace the port number between 1024 to 65535. Source can parallel.
3.  callback the provided url to report the real port to listen
4.  user push streaming data to netty server, then collect the data to flink

### DIFF
--- a/flink-connector-netty/pom.xml
+++ b/flink-connector-netty/pom.xml
@@ -40,11 +40,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.flink</groupId>
-      <artifactId>flink-table_${scala.binary.version}</artifactId>
-      <version>${flink.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.flink</groupId>
       <artifactId>flink-streaming-scala_${scala.binary.version}</artifactId>
       <version>${flink.version}</version>
     </dependency>
@@ -73,6 +68,12 @@
       <version>4.5.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.flink</groupId>
+      <artifactId>flink-table_${scala.binary.version}</artifactId>
+      <version>${flink.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
@@ -80,7 +81,6 @@
       <plugin>
         <groupId>net.alchim31.maven</groupId>
         <artifactId>scala-maven-plugin</artifactId>
-        <version>3.2.2</version>
       </plugin>
     </plugins>
   </build>
--- a/flink-connector-netty/pom.xml
+++ b/flink-connector-netty/pom.xml
@@ -1,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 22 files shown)


flink-connector-netty/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+      <version>4.1.5.Final</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.flink</groupId>
+      <artifactId>flink-table_${scala.binary.version}</artifactId>
+      <version>${flink.version}</version>
+    </dependency>
```
rmetzger: I don't think its a good idea to add the Table API here as a dependency just for one connector example.

Flink Table could grow quite big in the future, and users of the connector will not execute a single line of the table API.

flink-connector-netty/src/main/scala/org/apache/flink/streaming/connectors/netty/example/NettyUtil.scala
```diff
@@ -0,0 +1,160 @@
...
+        val result = startService(tryPort)
+        logger.info(s"Successfully started service$serviceString, result:$result.")
+        return result
+      } catch {
+        case e: Exception if isBindCollision(e) =>
+          if (offset >= maxRetries) {
+            val exceptionMessage = s"${e.getMessage}: Service$serviceString failed after " +
+              s"$maxRetries retries! Consider explicitly setting the appropriate port for the " +
+              s"service$serviceString (for example spark.ui.port for SparkUI) to an available " +
+              "port or increasing spark.port.maxRetries."
```
rmetzger: The exception message doesn't seem to be relevant to Flink.

In general, it seems that the code here seems to be copied from Apache Spark: https://github.com/apache/spark/blob/39755169fb5bb07332eef263b4c18ede1528812d/core/src/main/scala/org/apache/spark/util/Utils.scala#L2172

Can you add comments to the code copied from other projects?

shijinkui: @rmetzger OK, added

shijinkui: @rmetzger I've added example to TcpReceiverSource and HttpReceiverSource. Also there are two full example in test package `TcpSourceExample` and `StreamSqlExample`

shijinkui:
@rmetzger any problem?

rmetzger:
Thanks a lot for the updates. I was traveling the last few weeks, that's why I didn't have time to review your changes earlier.

I'll merge the PR.

ckadner:
Sadly we did not have *Scalatest* enabled at the time this PR was reviewed, so we missed adding automated unit tests.

I opened [BAHIR-113: Flink Netty connector missing automated unit tests](https://issues.apache.org/jira/browse/BAHIR-113) to keep track of that.

@shijinkui -- would you be willing to take that on and open a PR for  [BAHIR-113](https://issues.apache.org/jira/browse/BAHIR-113)?

PR #9, opened 2017-01-24
[BAHIR-85] move getCommandDescription to invoke method

atharvai:
REF: https://issues.apache.org/jira/browse/BAHIR-85

Moves `getCommandDescription()` call to inside `invoke()` to allow changes to command key. 

Questions:
1. Do we need to upgrade version?


rmetzger:
I don't think that this change will allow you to update the key while the redis sink is running.

ckadner:
@atharvai @rmetzger -- are you still working on this? 

- should it be marked as WIP?
- is it waiting for review?
- otherwise close it?

Thanks!

rmetzger:
I don't fully understand the purpose of this change. I propose to close the pull request and discuss the implementation in the JIRA first before opening another PR.

PR #10, opened 2017-02-28
[BAHIR-93] Correct all version strings to be 1.0-SNAPSHOT

tzulitai:
According to the discussion in https://www.mail-archive.com/dev@bahir.apache.org/msg00448.html, we will be following a "major.minor" versioning format for bahir-flink, starting from `1.0`.
Therefore, the current state being pre-release of 1.0, the correct version in master should be `1.0-SNAPSHOT`.

This PR corrects all version string occurrences in `bahir-flink`, to prepare it for the upcoming 1.0 release.


rmetzger:
Thank you for opening a pull request.
The change looks good +1 to merge.

PR #11, opened 2017-03-01
[BAHIR-91] Upgrade Flink version to 1.2.0

tzulitai:
This required some Flink state API updates in the tests. We probably should keep an extra eye on these kind of Flink API changes; the source APIs may be stable, but API changes like these can also break code in bahir-flink. (I suggest a new major version for bahir-flink every time a API break happens from Flink, and as soon as that happens for the first time from now on, we start maintaining a compatibility matrix for bahir-flink).

Note about the additional `maven-bundle-plugin` in the Redis connector:
After the 1.2.0 upgrade, the project had a new dependency to Apacheds JDBM 2.0.0-M2, which packages using the "bundle" type. The plugin helps recognizes the type when building.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
     <log4j.version>1.2.17</log4j.version>
 
     <!-- Flink version -->
-    <flink.version>1.1.1</flink.version>
+    <flink.version>1.2.0</flink.version>
 
     <PermGen>64m</PermGen>
     <MaxPermGen>512m</MaxPermGen>
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,7 @@
 language: java
 
 env:
-#  - FLINK_VERSION="1.1.0"
-  - FLINK_VERSION="1.1.1"
+  - FLINK_VERSION="1.2.0"
 
 jdk:
   - oraclejdk8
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -90,7 +90,7 @@
     <log4j.version>1.2.17</log4j.version>
 
     <!-- Spark version -->
```
tzulitai: minor note: I changed this line to `Flink version` in #10, so not changing it in this PR to avoid conflicts.

rmetzger:
I think the travis tests will fail because they are overwriting the Flink version.
I have a constant there for the Flink version. My initial plan was to support multiple Flink versions at the same time.
Do you think there is an easy way to still support 1.1. and 1.2 with the Bahir connectors?

tzulitai:
@rmetzger I don't think there's an easy way to do this, unless we try to make the affected connectors backwards compatible.

Just rebased on {{master}} to resolve conflicts.

rmetzger:
The travis build failed, probably because of the version override from the travis.yml.

tzulitai:
Ah yes, I forgot about that. I'll change that to `1.2.0`.

rmetzger:
Looks good. Merging change.

PR #13, opened 2017-03-14
[BAHIR-95] Add ZREM to Redis commands

ariskk:
Adds ZREM.
Unless there is a really good reason not to, i intend to add a few others as well,

### DIFF
--- a/flink-connector-redis/README.md
+++ b/flink-connector-redis/README.md
@@ -141,6 +141,9 @@ This section gives a description of all the available data types and what Redis 
         </tr>
         <tr>
             <td>SORTED_SET</td><td><a href="http://redis.io/commands/zadd">ZADD</a></td>
-        </tr>                
+        </tr>
+        <tr>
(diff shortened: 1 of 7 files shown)


rmetzger:
I think the reason why `ZREM` was not added is because this is the `RedisSink`, which implies that you are writing data to redis, not removing it.
I'm not sure if it makes sense semantically. Do you have a use case for removing data in the sink?

The change looks good overall.
Maybe you could also update the `README.md` file with the new command?

ariskk:
In our use case, both additions and deletions are processed via flink.
Think of a graph for example, that you store nodes as keys and their edges as sorted sets (eg sorted on date created). Assuming this is dynamic, edges may be added, but may also be removed.
I agree it wouldn't make much sense to delete keys, but the ability to add and remove elements from sets is important.
We are using our own Redis Sink up to now, which supports this, but we are thinking to migrate to the official one.

rmetzger:
Thank you for updating the PR.

I'll merge it.

I'm planning to create the first bahir-flink release soon. Would you like to first add more commands to the connector, or is it good to go for a first release?

PR #14, opened 2017-03-24
[BAHIR-96] Add release util scripts for Bahir Flink

tzulitai:
The release scripts are basically copied from the `bahir` repository, with adjustments to the `release-build.sh` so that the `bahir-flink` urls are used, and usage examples follow Bahir Flink's `major.minor` versioning format.

A dry run of the `release-build.sh` script does not result in any errors.

### DIFF
--- a/dev/release-build.sh
+++ b/dev/release-build.sh
@@ -264,7 +264,7 @@ if [[ "$RELEASE_PREPARE" == "true" ]]; then
         cd .. #exit $RELEASE_VERSION-$RELEASE_RC/
 
         svn add $RELEASE_VERSION-$RELEASE_RC/
-        svn ci -m"Apache Bahir Flink Extensions $RELEASE_VERSION-$RELEASE_RC"
+        svn ci -m"Apache Bahir extensions for Apache Flink $RELEASE_VERSION-$RELEASE_RC"
     fi
 
 
--- a/dev/release-build.sh
+++ b/dev/release-build.sh
@@ -29,7 +29,7 @@ usage: release-build.sh [--release-prepare | --release-publish | --release-snaps
 DESCRIPTION
 
 Use maven infrastructure to create a project release package and publish
-to staging release location (https://dist.apache.org/repos/dist/dev/bahir-flink)
+to staging release location (https://dist.apache.org/repos/dist/dev/bahir/bahir-flink)
 and maven staging release repository.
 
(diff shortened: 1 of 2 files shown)


dev/release-build.sh
```diff
@@ -0,0 +1,330 @@
...
+
+if [ -z "$RELEASE_RC" ]; then
+  RELEASE_RC="rc1"
+fi
+
+if [ -z "$RELEASE_TAG" ]; then
+  RELEASE_TAG="v$RELEASE_VERSION-$RELEASE_RC"
+fi
+
+RELEASE_STAGING_LOCATION="https://dist.apache.org/repos/dist/dev/bahir-flink/"
```
lresende: The release svn location has the pattern of one folder per project, so I would say we would need to have something like bahir/bahir-flink, and probably also change bahir to be bahir/bahir-spark

dev/release-build.sh
```diff
@@ -0,0 +1,330 @@
...
+        cp bahir-flink/distribution/target/*.zip    svn-bahir-flink/$RELEASE_VERSION-$RELEASE_RC/
+
+        cd svn-bahir-flink/$RELEASE_VERSION-$RELEASE_RC/
+        for i in *.zip *.gz; do gpg --output $i.asc --detach-sig --armor $i; done
+        for i in *.zip *.gz; do openssl md5 -hex $i | sed 's/MD5(\([^)]*\))= \([0-9a-f]*\)/\2 *\1/' > $i.md5; done
+
+        cd .. #exit $RELEASE_VERSION-$RELEASE_RC/
+
+        svn add $RELEASE_VERSION-$RELEASE_RC/
+        svn ci -m"Apache Bahir Flink Extensions $RELEASE_VERSION-$RELEASE_RC"
```
lresende: How about change this to:
"Apache Bahir extensions for Apache Flink ...."

Which will make things following the same pattern as the download page (recent suggested changes)

tzulitai: That sounds better, will change this!

dev/release-build.sh
tzulitai: Ah, I see. +1 to this pattern, changing.

lresende:
Thanks for the quick update. 

LGTM, will merge if there are no more comments.

tzulitai:
any updates for this?

PR #21, opened 2017-08-28
[BAHIR-134] Add InfluxDB sink for flink stream

yew1eb:
add InfluxDBSink for flink stream

## Verifying this change
Add example `InfluxDBSinkExample`
Runing environment:
flink version 1.3.0
influxdb version: 1.3.4
influxdb-java(influxdb client) version: 2.7(Compatible with InfluxDB version 0.9 ~ 1.3.x, but requirements  JDK 1.8+)

It works well.

### DIFF
--- a/flink-connector-influxdb/src/main/java/org/apache/flink/streaming/connectors/influxdb/InfluxDBSink.java
+++ b/flink-connector-influxdb/src/main/java/org/apache/flink/streaming/connectors/influxdb/InfluxDBSink.java
@@ -33,31 +33,16 @@ import java.util.concurrent.TimeUnit;
  */
 public class InfluxDBSink extends RichSinkFunction<InfluxDBPoint> {
 
-    private transient InfluxDB influxDB = null;
-    private final String dbName;
-    private final String username;
-    private final String password;
-    private final String host;
-    private boolean batchEnabled = true;
+    private transient InfluxDB influxDBClient;
+    private final InfluxDBConfig influxDBConfig;
 
     /**
      * Creates a new {@link InfluxDBSink} that connects to the InfluxDB server.
      *
-     * @param host     the url to connect to.
-     * @param username the username which is used to authorize against the influxDB instance.
-     * @param password the password for the username which is used to authorize against the influxDB instance.
-     * @param dbName   the database to write to.
+     * @param influxDBConfig The configuration of {@link InfluxDBConfig}
      */
-    public InfluxDBSink(String host, String username, String password, String dbName) {
-        this.host = Preconditions.checkNotNull(host, "host can not be null");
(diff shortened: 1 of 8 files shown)


flink-connector-influxdb/src/main/java/org/apache/flink/streaming/connectors/influxdb/InfluxDBSink.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Initializes the connection to InfluxDB by either cluster or sentinels or single server.
+     */
+    @Override
+    public void open(Configuration parameters) throws Exception {
+        super.open(parameters);
+
+        influxDB = InfluxDBFactory.connect(host, username, password);
+        if (!influxDB.databaseExists(dbName)) {
+            influxDB.createDatabase(dbName);
```
rmetzger: Maybe for the future: I would actually log an info message, stating that you've created the database.
I think its a good if code always tells the user when it is doing magic.

flink-connector-influxdb/src/main/java/org/apache/flink/streaming/connectors/influxdb/InfluxDBSink.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        if (batchEnabled) {
+            // Flush every 2000 Points, at least every 100ms
+            influxDB.enableBatch(2000, 100, TimeUnit.MILLISECONDS);
```
rmetzger: Ideally we'll make this configurable.

rmetzger:
Overall, I like the change.
The only problem is that it breaks the build because of missing Java 7 support.

I guess we need to decide whether we want to drop java 7 support.

mridulm:
As of spark 2.2, java 7 support has been removed.
I think flink also requires java 8, right ?

Perhaps we can do the same for bahir

rmetzger:
Flink 1.3, the current release has still Java 7 support. But starting from Flink 1.4, Java 7 will be dropped.
I'll start a quick discussion on the dev@ list, to make sure nobody disagrees.

flink-connector-influxdb/src/main/java/org/apache/flink/streaming/connectors/influxdb/InfluxDBSink.java
yew1eb: Thank you for review.
I will use `throw RuntimeException` instead of helping the user to create. 
Because `open` method will being execute by all parallel sink tasks.

yew1eb:
Hi @rmetzger, thanks for your review. I have updated the PR.
Best,
Hai Zhou
:beers:

rmetzger:
Thanks. I'll merge the change.

PR #25, opened 2017-12-05
add redis sink checkpoint

buptljy:
It seems that the redis sink is not very suitable for some specific circumstances.
For example, I want to do a word count and I enable the checkpoint machenism, it seems to be wrong if I start my program from checkpoint when my program fails, because datas that already sent will be sent to redisSink again.
I don't know if it is acceptable for you, so I write a simple version with a flushOnCheckpoint option. I have a further improvement idea that we can use redis pipeline if we are going to store datas into redis when snapshot. 
I will optimize it if this idea looks good to you.


rmetzger:
Thanks a lot for the pull request. Integrating checkpointing with external connectors is usually a tricky topic. Before we can merge this change, I'd like to fully understand the problem first.

Does the Jedis library buffer writes internally? This would interfere with the checkpointing (like its the case with Flink's Kafka connector), but I don't think that jedis is buffering.

PR #28, opened 2018-06-16
Correcting connector version in README.md

antonrud:
Otherwise the dependency could not be resolved.


lresende:
Should you also look into the rest of the code for any other references of 1.0-SNAPSHOT? 
https://github.com/apache/bahir-flink/search?q=1.0-SNAPSHOT&unscoped_q=1.0-SNAPSHOT

PR #32, opened 2018-08-10
[BAHIR-176] update to flink 1.6.0

eskabetxe:


eskabetxe:
@lresende could you check this. thanks

PR #42, opened 2019-02-01
Enhanced create RpcClient and FlumeSink

ambition119:
Create more RpcClient type and Support FlumeSink batch write. 

 [Flink PR 7618](https://github.com/apache/flink/pull/7618)


eskabetxe:
thanks for the PR, its a great improvement of flume connector.
but, why have you created a new sink instead of improving the existing one?
mainly because the existing one is very basic and easily adaptable to your version

lresende:
+1 from @eskabetxe comments above, duplicating a bunch of files as V1, V2, Vn will not scale and will make it hard to maintain. Please describe if you have specific requirements where you can't enhance the existing version of the sink.

ambition119:
@eskabetxe @lresende  I improved the existing FlumeSink and removed the code content of v2.
The improvements are:
Richly create the type of RpcClient, default, thrift, etc.
Flume Event's abstract interface creates an implementation interface that replaces the previous code 
 like :
```java
    public boolean sendData(String data) 

    public boolean sendData(byte[] data) 

    private boolean sendData(Event event) 
```

eskabetxe:
@lresende  LGTM

flink-connector-flume/src/main/java/org/apache/flink/streaming/connectors/flume/FlumeSink.java
```diff
@@ -1,64 +1,150 @@
...
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
```
lresende: Please don't change the license headers...

ambition119: ok

PR #43, opened 2019-02-05
[BAHIR-193]: facilitate scala version change

eskabetxe:
@lresende can you check this change


lresende:
With your proposed change, how do you build different versions of Scala? Also, for a downstream consumer, how will it use 2.11 versus 2.12 version of a dependency if both will point to ONE parent pom that has ONE definition of scala.binary.version ?

eskabetxe:
hi @lresende, this approach its the same as flink and other projects
the parent really is for compilation, if you change de scala version on parent, the modules adequate their naming with that change

lresende:
Flink seems to use something similar, but most of other projects I have worked before (e.g. Apache Spark) use the current approach used in Bahir. Also, it seems scala best practices to have all artifacts published with the version as part of the artifact name (e.g. _2.11 or _2.12).

Now, back to building, my understanding is that, when you have the parent pom released, it becomes a static file, and the pre-defined scala.version is also static. So, when when a project add a dependency on <flink-connector-kudu_2.12> it will have an transient dependency to <flink_parent> that has a `scala.binary.version` defined as 2.11, unless it provides the same infrastructure of profiles in their build to override the property, no ? 

Please let me know what you think, and in the meantime I will play with it a little more.

eskabetxe:
The artifacts will have de correctly _2.11 or _2.12 as you activate the right profile , and the dependencies will be created with the right scala version.
What you tell of parent, that could have a bad property definition its true, but you have compile the jar with the right version.
Maybe we have to change something when creating the releases, but if you want to change from one version of scala to another it is easier.

lresende:
With the recent changes on #46, does this become obsolete?

PR #50, opened 2019-03-22
[BAHIR-202] Improve KuduSink throughput by using async FlushMode

SuXingLee:
change is that:
1. Default , KuduSink processing message one by one without checkpoint.
2. when we enable checkpoint , We improve kuduSink throughput by using ```FlushMode.AUTO_FLUSH_BACKGROUND``` , and using checkpoint to ensure at-least-once.

### DIFF
--- a/flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/KuduSink.java
+++ b/flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/KuduSink.java
@@ -17,20 +17,27 @@
 package org.apache.flink.streaming.connectors.kudu;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.runtime.state.FunctionInitializationContext;
+import org.apache.flink.runtime.state.FunctionSnapshotContext;
+import org.apache.flink.streaming.api.checkpoint.CheckpointedFunction;
 import org.apache.flink.streaming.api.functions.sink.RichSinkFunction;
+import org.apache.flink.streaming.api.operators.StreamingRuntimeContext;
 import org.apache.flink.streaming.connectors.kudu.connector.KuduConnector;
 import org.apache.flink.streaming.connectors.kudu.connector.KuduRow;
 import org.apache.flink.streaming.connectors.kudu.connector.KuduTableInfo;
 import org.apache.flink.streaming.connectors.kudu.serde.KuduSerialization;
 import org.apache.flink.util.Preconditions;
+import org.apache.kudu.client.SessionConfiguration.FlushMode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 
-public class KuduSink<OUT> extends RichSinkFunction<OUT> {
+public class KuduSink<OUT> extends RichSinkFunction<OUT> implements CheckpointedFunction {
 
-    private static final Logger LOG = LoggerFactory.getLogger(KuduOutputFormat.class);
+    private static final long serialVersionUID = 1L;
+
+	private static final Logger LOG = LoggerFactory.getLogger(KuduOutputFormat.class);
 
     private String kuduMasters;
     private KuduTableInfo tableInfo;
@@ -79,9 +86,12 @@ public class KuduSink<OUT> extends RichSinkFunction<OUT> {
 
     @Override
     public void open(Configuration parameters) throws IOException {
-        if (connector != null) return;
-        connector = new KuduConnector(kuduMasters, tableInfo, consistency, writeMode);
-        serializer.withSchema(tableInfo.getSchema());
+        if (this.connector != null) return;
+        FlushMode flushMode = ((StreamingRuntimeContext) getRuntimeContext()).isCheckpointingEnabled() ?
+        		FlushMode.AUTO_FLUSH_BACKGROUND :FlushMode.AUTO_FLUSH_SYNC;
+        this.connector = new KuduConnector(kuduMasters, tableInfo, consistency, writeMode, flushMode);
+        this.serializer.withSchema(tableInfo.getSchema());
+        
     }
 
     @Override
@@ -102,5 +112,17 @@ public class KuduSink<OUT> extends RichSinkFunction<OUT> {
         } catch (Exception e) {
             throw new IOException(e.getLocalizedMessage(), e);
         }
+    }
+
+	@Override
+    public void initializeState(FunctionInitializationContext context) throws Exception {
+    }
+
+	@Override
+    public void snapshotState(FunctionSnapshotContext context) throws Exception {
+        if (LOG.isDebugEnabled()) {
+        	LOG.debug("Snapshotting state {} ...", context.getCheckpointId());
+        }
+		this.connector.flush();
     }
 }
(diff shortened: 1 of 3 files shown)


SuXingLee:
@eskabetxe  clould you check this

flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/connector/KuduConnector.java
```diff
@@ -133,8 +136,15 @@ public void close() throws Exception {
 
         if (client == null) return;
         client.close();
+        
+        if (session == null) return;
```
eskabetxe: you should close session first and then client no?

flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/KuduSink.java
```diff
@@ -79,9 +86,12 @@ public KuduSink(String kuduMasters, KuduTableInfo tableInfo, KuduSerialization<O
 
     @Override
     public void open(Configuration parameters) throws IOException {
-        if (connector != null) return;
-        connector = new KuduConnector(kuduMasters, tableInfo, consistency, writeMode);
-        serializer.withSchema(tableInfo.getSchema());
+        if (this.connector != null) return;
+        FlushMode flushMode = ((StreamingRuntimeContext) getRuntimeContext()).isCheckpointingEnabled() ?
```
eskabetxe: I dont know if AUTO_FLUSH_BACKGROUND show be used here..

From Kudu documentacion
Note: The AUTO_FLUSH_BACKGROUND mode may result in out-of-order writes to Kudu. This is because in this mode multiple write operations may be sent to the server in parallel. See KUDU-1767 for more information.

eskabetxe:
@SuXingLee you have 12 Checkstyle violations, could you check this

flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/connector/KuduConnector.java
SuXingLee: thanks for your review, have been fixed

flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/KuduSink.java
SuXingLee: Now, the logic is :
-  If flink checkpoint is disable,synchronously write data to kudu.
-  If flink checkpoint is enable, asynchronously write data to kudu by default.
-  user also can change to sync by explicitly calling```KuduSink.withSyncFlushMode()``` when initializing KuduSink.  

the out-of-order tip is writed in ```KuduSink.withSyncFlushMode()``` comments, if someone want to orderly write data to kudu when checkpoint is enable, he can explicitly calling```KuduSink.withSyncFlushMode()``` .

SuXingLee:
> @SuXingLee you have 12 Checkstyle violations, could you check this

have been fixed

flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/KuduSink.java
eskabetxe: I have doubts if we should not merge these methods
-> withEventualConsistency - withAsyncFlushMode
-> withStrongConsistency - withSyncFlushMode

SuXingLee: you mean to unite consistency's config with flushmode's config ? 
-  ```ASYNC_MODE ( EVENTUAL + AUTO_FLUSH_BACKGROUND )```
-  ```SYNC_MODE ( STRONG + AUTO_FLUSH_SYNC )```

eskabetxe: yes.
its the same no?
the eventual consistence use async session and callback for response
the strong consistence force a response.join() to obtain the response (converting the call to a synchronous call)

if i understood correctly the auto_flush its doing the same no?

SuXingLee: In my understanding they are not equivalent.
- ```EVENTUAL``` and ```STRONG ``` mostly means out-of-order or ordered send message to Kudu.
- ```AUTO_FLUSH_BACKGROUND``` and ```AUTO_FLUSH_SYNC``` mostly means batch or single send message to Kudu.

```response.join()``` is Synchronously waits until this Deferred is called back.  If this Deferred already completed, this method returns (or throws) immediately. Otherwise, the current thread will be blocked and will wait until the Deferred is called back.
the  ```response.join()```  is only useful when you want to keep strong consistence In operation interleaving , eg: A opeartion and B opeartion need to be executed in order, when we use ```session.apply(operation)```  each opeartion will use client to send Rpc Request . in this step, there are sync, but maybe B opeartion is execute fast than A in kudu server，and we first got B opeartion 's callback. So it don't ensure consistence, only if we use ```response.join()```  to keep 
 waiting for A opeartion's callback before execute B opeartion.
But, whether we use ```response.join()``` or not. In default FlushMode ```session.apply(operation)`` is execute operation one by one , not batch. if we want to use buffer to impove throughput ,we can use another FlushMode like ```AUTO_FLUSH_BACKGROUND``` , and using checkpoint to ensure at-least-once.

eskabetxe: so the autofush will only make sense in eventual consistence no?

SuXingLee: the eventual consistence only make sense in AUTO_FLUSH_SYNC FlushMode.

eskabetxe: and strong consistence is ok with any flush mode?

eskabetxe:
@lresende LGTM

flink-connector-kudu/src/main/java/org/apache/flink/streaming/connectors/kudu/KuduSink.java
SuXingLee: - ```AUTO_FLUSH_BACKGROUND``` don't support  strong consistence.
- ```MANUAL_FLUSH``` support  strong consistence by wait for one flush() and join() before triggering another flush. but in this mode, flush data only depend on checkpoint trigger. is it suitable in flink sink? should we add another strategy(count、size) ? we shoud do more in another PR.

eskabetxe: yes, maybe we have to rethink this.
we are forcing to chose strong or eventual consistence, but seems that some flush modes are not ok with any of these options..

could your create a new issue so we can discuss the best option

SuXingLee: ok, i will create a another issue to discuss it atfer a few days later.
And thanks for your review again.

PR #51, opened 2019-04-01
[BAHIR-177] Fixed state recovery for Flink and fixed size of the reco…

Wosin:
Two issues are meant to be fixed in this PR:
1) As described in BAHIR-177 currently the state recovery of Bahir operators depends on randomly generated IDs, which basically makes it impossible to recover state properly. The chagne has been done, so that the `outStreamId` is used instead of random names.
2) The size of the queue recovered in `restoreQueuerState()` was equal to the actual size (number of elements) of the snapshot queue. If the queue was empty, the method would try to create queue with the size 0, which is currently forbidden for the `PriorityQueue` in Java.


eskabetxe:
@lresende LGTM

PR #52, opened 2019-04-11
[BAHIR-204] [activemq] ActiveMQ Source only emits previously unproce…

knaufk:
ActiveMQ Source only emits previously unprocessed records now. Previously, it was possible that this source duplicates records if the Flink job failed between snapshotState and notifyOnCheckpointComplete


eskabetxe:
@lresende LGTM

rmetzger:
Merging this change ...

rmetzger:
Thank you!

PR #57, opened 2019-05-25
[BAHIR-205] Support configure password for redis cluster

liketic:
Upgrade jedis to 2.9.0 and add password for redis cluster sink.


flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/common/config/JedisClusterConfigTest.java
```diff
@@ -46,4 +48,20 @@ public void shouldThrowIllegalArgumentExceptionIfNodeValuesAreEmpty(){
             .setNodes(set)
             .build();
     }
+
+    @Test
+    public void shouldSetPasswordSuccessfully() {
```
eskabetxe: could you add a test with no password setted.

liketic: I added the test.

eskabetxe:
@lresende LGTM

PR #58, opened 2019-06-07
[BAHIR-207] added tests for scala 2.12 on travis

eskabetxe:


eskabetxe:
@lresende could you check

PR #59, opened 2019-06-07
[BAHIR-207] added tests for scala 2.12 on travis

eskabetxe:


lresende:
What are the benefits of breaking the builds per project?

pom.xml
```diff
@@ -703,7 +703,7 @@
       <id>scala-2.11</id>
       <activation>
         <property>
-          <name>!scala-2.12</name>
+          <name>scala-2.11</name>
         </property>
```
lresende: If we don't specify a profile on the command line, does the build pickup a default one and build it properly?

eskabetxe:
> What are the benefits of breaking the builds per project?

none.. travis was giving me a bad day..
its fixed

pom.xml
eskabetxe: on travis is mandarory because of change-scala-version.sh

on local machine it works as scala version is defined on properties.. (version 2.11)

eskabetxe:
solved the problems with travis..
akka connector and siddhi library have a lot of warnings and logs

eskabetxe:
@lresende could you check this

PR #60, opened 2019-07-01
BAHIR-85: make it possible to change additional key without restarting

tonvanbart:
We have a use case where we want to sink data to Redis as hashes, but have the hashes stored under different keys which are also extracted from the data. This change makes this possible in a backward compatible manner (see my comment in https://issues.apache.org/jira/browse/BAHIR-85 for details).

### DIFF
--- a/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/RedisSink.java
+++ b/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/RedisSink.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


marcosschroh:
Hi,

Any update in this? :smiley:

eskabetxe:
@lresende LGTM

marcosschroh:
hi @lresende ,

When the a new version is going to be released with the new change?

lresende:
I don't think I can do it before OSCON which is next week, any volunteers ? @eskabetxe or @lukasz-antoniak

Zentopia:
Any updates on this?  #We need it.

lresende:
Any volunteers to help with a release ?

eskabetxe:
@lresende I help you

PR #62, opened 2019-07-03
[BAHIR-210] upgrade flink to 1.8.1

eskabetxe:
upgrade flink to 1.8.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,8 +49,8 @@ before_install:
 script:
   - if [[ $DOCKER == "true" ]]; then
     docker-compose -f "$PROJECT/dockers/docker-compose.yml" up -d;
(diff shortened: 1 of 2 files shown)


eskabetxe:
@lresende could you check

lresende:
@eskabetxe could you check the build failures, it might be related to Travis load, but it usually passes so it might also be a side effect of the upgrade?

eskabetxe:
this is the error 
"The job exceeded the maximum log length, and has been terminated."

i have pending to check all the warnings..
i will see if there are some easy error to solve

eskabetxe:
@lresende i added a -quiet option to mvn command.. 
that way only logs the errors...

PR #64, opened 2019-08-13
[BAHIR-214]: improve speed and solve issues on eventual consistence

eskabetxe:
- resolve eventual consistence issues
- improve speed special on eventual consistence stream
- actualized Readme


eskabetxe:
@lresende  could you check

eskabetxe:
@lresende this is falling for the same reason BAHIR-210 fails..
this is the error "The job exceeded the maximum log length, and has been terminated."

its resolved on that PR, could yo merge that first?

eskabetxe:
@lresende the build its fixed.. could you check

PR #65, opened 2019-08-26
[BAHIR-215]: bump flink version to 1.9.0

eskabetxe:
bump flink version to 1.9.0


eskabetxe:
@lresende could you check this

PR #72, opened 2019-11-19
[BAHIR-220] Add redis descriptor to make redis connection as a table

hzyuemeng1:
Currently, for Flink-1.9.0, we can use the catalog to store our stream table source and sink
for Redis connector, it should exist a Redis table sink so we can register it to catalog, and use Redis as a table in SQL environment

```
Redis redis = new Redis()
                .mode(RedisVadidator.REDIS_CLUSTER)
                .command(RedisCommand.INCRBY_EX.name())
                .ttl(100000)
                .property(RedisVadidator.REDIS_NODES, REDIS_HOST+ ":" + REDIS_PORT);
tableEnvironment
                .connect(redis).withSchema(new Schema()
                .field("k", TypeInformation.of(String.class))
                .field("v", TypeInformation.of(Long.class)))
                .registerTableSink("redis");
tableEnvironment.sqlUpdate("insert into redis select k, v from t1");
env.execute("Test Redis Table");
```

### DIFF
--- a/flink-connector-redis/pom.xml
+++ b/flink-connector-redis/pom.xml
@@ -43,6 +43,11 @@ under the License.
             <artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
             <version>${flink.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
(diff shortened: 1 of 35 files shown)


eskabetxe:
@lresende LGTM

PR #77, opened 2020-04-01
[hotfix]Fix SetMapper getAdditionalTTL throw NPE

Tartarus0zm:
ttl is Integer, but `getTtl()` return is int, SET command has not ttl


Tartarus0zm:
@hzyuemeng1  please take a look，thx

PR #82, opened 2020-05-21
Add batch table env support and filter push down

sebastianliu:
It was glad to see that community's Flink-kudu connector has been reworked recently. The table api has been supported which enable better interaction between Flink SQL and Kudu. However, there is no support for flink batch SQL currently, include filter pushing down. In this PR, I change KuduTableSource to inherit from InputFormatTableSource to support both streaming SQL and Batch SQL at the same time. In order to reduce unnecessary data transmission, I also added the filter push down to the KuduTableSource.

### DIFF
--- a/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableFactory.java
+++ b/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableFactory.java
@@ -138,7 +138,7 @@ public class KuduTableFactory implements TableSourceFactory<Row>, TableSinkFacto
         KuduReaderConfig.Builder configBuilder = KuduReaderConfig.Builder
                 .setMasters(masterAddresses);
 
-        return new KuduTableSource(configBuilder, tableInfo, physicalSchema, null);
+        return new KuduTableSource(configBuilder, tableInfo, physicalSchema, null, null);
     }
 
     @Override
--- a/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
+++ b/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
@@ -19,12 +19,15 @@ package org.apache.flink.connectors.kudu.table;
 
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.connectors.kudu.batch.KuduRowInputFormat;
+import org.apache.flink.connectors.kudu.connector.KuduFilterInfo;
 import org.apache.flink.connectors.kudu.connector.KuduTableInfo;
 import org.apache.flink.connectors.kudu.connector.reader.KuduReaderConfig;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.DataTypes;
 import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.table.expressions.Expression;
+import org.apache.flink.table.sources.FilterableTableSource;
 import org.apache.flink.table.sources.LimitableTableSource;
 import org.apache.flink.table.sources.ProjectableTableSource;
 import org.apache.flink.table.sources.StreamTableSource;
@@ -35,35 +38,73 @@ import org.apache.flink.table.types.utils.TypeConversions;
 import org.apache.flink.types.Row;
 
 import org.apache.flink.shaded.guava18.com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.ListIterator;
+import java.util.Optional;
 
-public class KuduTableSource implements StreamTableSource<Row>, LimitableTableSource<Row>, ProjectableTableSource<Row> {
+import static org.apache.flink.connectors.kudu.table.utils.KuduTableUtils.toKuduFilterInfo;
+
+public class KuduTableSource implements StreamTableSource<Row>,
+    LimitableTableSource<Row>, ProjectableTableSource<Row>, FilterableTableSource<Row> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KuduTableSource.class);
 
     private final KuduReaderConfig.Builder configBuilder;
     private final KuduTableInfo tableInfo;
     private final TableSchema flinkSchema;
     private final String[] projectedFields;
+    // predicate expression to apply
+    @Nullable
+    private final List<KuduFilterInfo> predicates;
+    private boolean isFilterPushedDown;
 
-    public KuduTableSource(KuduReaderConfig.Builder configBuilder, KuduTableInfo tableInfo, TableSchema flinkSchema, String[] projectedFields) {
+    private KuduRowInputFormat kuduRowInputFormat;
+
+    public KuduTableSource(KuduReaderConfig.Builder configBuilder, KuduTableInfo tableInfo,
+        TableSchema flinkSchema, List<KuduFilterInfo> predicates, String[] projectedFields) {
         this.configBuilder = configBuilder;
         this.tableInfo = tableInfo;
         this.flinkSchema = flinkSchema;
+        this.predicates = predicates;
         this.projectedFields = projectedFields;
+        if (predicates != null && predicates.size() != 0) {
+            this.isFilterPushedDown = true;
+        }
+        this.kuduRowInputFormat = new KuduRowInputFormat(configBuilder.build(), tableInfo,
+            predicates == null ? Collections.emptyList() : predicates,
+            projectedFields == null ? null : Lists.newArrayList(projectedFields));
+    }
+
+    @Override
+    public boolean isBounded() {
+        return true;
     }
 
     @Override
     public DataStream<Row> getDataStream(StreamExecutionEnvironment env) {
-
-        KuduRowInputFormat inputFormat = new KuduRowInputFormat(configBuilder.build(), tableInfo, projectedFields == null ? null : Lists.newArrayList(projectedFields));
-
-        return env.createInput(inputFormat, (TypeInformation<Row>) TypeConversions.fromDataTypeToLegacyInfo(getProducedDataType())).name(explainSource());
+        KuduRowInputFormat inputFormat = new KuduRowInputFormat(configBuilder.build(), tableInfo,
+            predicates == null ? Collections.emptyList() : predicates,
+            projectedFields == null ? null : Lists.newArrayList(projectedFields));
+        return env.createInput(inputFormat,
+            (TypeInformation<Row>) TypeConversions.fromDataTypeToLegacyInfo(getProducedDataType()))
+            .name(explainSource());
     }
 
     @Override
     public TableSchema getTableSchema() {
         return flinkSchema;
+    }
+
+    @Override
+    public boolean isFilterPushedDown() {
+        return this.isFilterPushedDown;
     }
 
     @Override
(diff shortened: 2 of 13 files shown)


sebastianliu:
@lresende Appreciate for your time to review this PR.

lresende:
@gyfora do you want to take a look at this as well ?

gyfora:
Thanks @lresende , I can take a look today :)

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -109,12 +156,167 @@ public boolean isLimitPushedDown() {
...
+        predicates.clear();
+        predicates.addAll(unsupportedExpressions);
+        return new KuduTableSource(configBuilder, tableInfo, flinkSchema, kuduPredicates, projectedFields);
+    }
+
+    /**
+     * Converts Flink Expression to KuduFilterInfo.
+     */
+    @Nullable
+    private KuduFilterInfo toKuduFilterInfo(Expression predicate) {
```
gyfora, line 189: Could we move all the logic translating between Expression to KuduFilterInfo to a Utility class? That would leave the source cleaner.

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -109,12 +156,167 @@ public boolean isLimitPushedDown() {
...
+                    "KuduTable [{}].", pred, kuduPred, tableInfo.getName());
+                kuduPredicates.add(kuduPred);
+            } else {
+                unsupportedExpressions.add(pred);
+                LOG.info("Predicate [{}] could not be pushed into KuduFilterInfo for KuduTable [{}].",
+                    pred, tableInfo.getName());
+            }
+        }
+        // update list of Flink expressions to unsupported expressions
+        predicates.clear();
```
gyfora, line 180: Tha javadocs of applyPredicate specifies that we should remove the applicable expressions from the list only.
I know it's not a large difference but maybe it would be better to use an iterator and remove the expression if pushed down.

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -17,55 +17,106 @@
...
+import java.util.Set;
 
-public class KuduTableSource implements StreamTableSource<Row>, LimitableTableSource<Row>, ProjectableTableSource<Row> {
+public class KuduTableSource extends InputFormatTableSource<Row> implements
+    LimitableTableSource<Row>, ProjectableTableSource<Row>, FilterableTableSource<Row> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KuduTableSource.class);
+
+    @SuppressWarnings("unchecked")
+    private static final Set<BasicTypeInfo> VALID_LITERAL_TYPE = new HashSet() {{
```
gyfora, line 68: Should probably go to a Utility class with all the filter logic

flink-connector-kudu/src/test/java/org/apache/flink/connectors/kudu/table/KuduTableSourceTest.java
```diff
@@ -0,0 +1,73 @@
...
+        List<Row> result = dataSet.collect();
+        // check result
+        assertEquals(5, result.size());
+        assertEquals("1001,Java for dummies,Tan Ah Teck,11.11,11",
+            result.get(0).toString());
+        tableEnv.sqlUpdate("DROP TABLE books");
+    }
+
+    @Test
+    public void testScanWithProjectionAndFilter() throws Exception {
```
gyfora, line 61: With this test we cannot be sure that the filters were actually pushed down and it works correctly.

We should also add a unit test style test that validates the KuduFilterInfos created and the KuduTableSource directly.
Things like the applyPredicate or isFilterPushed down and all the rest.

The problem here is that we add a lot of different filtering logic that can easily break the output of SQL queries if incorrect, so we really want to make sure that it is tested.

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
sebastianliu: Fixed. This has been changed to use iterator. Thx for this reminder

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
sebastianliu: done

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
sebastianliu: Moved to KuduTableUtils

flink-connector-kudu/src/test/java/org/apache/flink/connectors/kudu/table/KuduTableSourceTest.java
sebastianliu: Good point. I have changed this test to be an integration test and added a new unit test style test for KuduTableSource. Thx

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableFactory.java
```diff
@@ -65,6 +65,7 @@
     public static final String KUDU_HASH_COLS = "kudu.hash-columns";
     public static final String KUDU_PRIMARY_KEY_COLS = "kudu.primary-key-columns";
     public static final String KUDU_REPLICAS = "kudu.replicas";
+    public static final String KUDU_IS_BOUNDED = "kudu.is-bounded";
```
gyfora, line 68: Why are we introducing the is-bounded flag? As far as I can tell this only affects the isBounded() method of the KuduTableSource and not the actual reading of the data.

The Kudu reading logic at the moment reads the table contents at the start of the job so it is always bounded, I think we should not touch this logic now. 

What do you think?

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -109,12 +147,41 @@ public boolean isLimitPushedDown() {
...
+        return new KuduTableSource(configBuilder, tableInfo, flinkSchema, bounded, predicates, fieldNames);
+    }
+
+    @Override
+    public TableSource<Row> applyPredicate(List<Expression> predicates) {
+        List<KuduFilterInfo> kuduPredicates = new ArrayList<>();
+        ListIterator<Expression> predicatesIter = predicates.listIterator();
+        while(predicatesIter.hasNext()) {
+            Expression predicate = predicatesIter.next();
+            KuduFilterInfo kuduPred = toKuduFilterInfo(predicate);
```
gyfora, line 159: Would be nicer to return Optional here but I leave this up to you

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -109,12 +147,41 @@ public boolean isLimitPushedDown() {
...
+            if (kuduPred != null) {
+                LOG.info("Predicate [{}] converted into KuduFilterInfo and pushed into " +
```
gyfora, line 161: Should we use debug logging here?

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -109,12 +147,41 @@ public boolean isLimitPushedDown() {
...
+                    "KuduTable [{}].", predicate, tableInfo.getName());
+                kuduPredicates.add(kuduPred);
+                predicatesIter.remove();
+            } else {
+                LOG.info("Predicate [{}] could not be pushed into KuduFilterInfo for KuduTable [{}].",
```
gyfora, line 166: debug log maybe?

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
```diff
@@ -109,12 +147,41 @@ public boolean isLimitPushedDown() {
...
-        return "KuduStreamTableSource[schema=" + Arrays.toString(getTableSchema().getFieldNames())
-                + (projectedFields != null ?", projectFields=" + Arrays.toString(projectedFields) + "]" : "]");
+        return "KuduTableSource[schema=" + Arrays.toString(getTableSchema().getFieldNames()) +
+            ", filter=" + predicateString() +
+            (projectedFields != null ?", projectFields=" + Arrays.toString(projectedFields) + "]" : "]");
+    }
+
+    private String predicateString() {
+        if (predicates == null || predicates.size() == 0) {
+            return "FALSE";
```
gyfora, line 182: Maybe instead of false, something like No Filters / predicates?

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
sebastianliu: Good suggestion, has changed to use `Optional` var

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableSource.java
sebastianliu: fixed

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableFactory.java
sebastianliu: For the `is-bounded` flag, it really just control the `isBounded()` method in 
`KuduTableSource` which inherit from `StreamTableSource`. The main reason I add this is because there may be some users who want to use this kudu connector in a Flink Batch SQL job. And this is usually used under the Blink Batch mode. I also added a few notes to the README file. In addition, I keep the default behavior of this `is-bounded` flag, which is false in `StreamTableSource`.  :)

sebastianliu:
> I had a few minor comments but it looks pretty good in general.
> 
> I also commented already on the is-bounded question, I think we should remove that logic from this PR as it does not introduce new functionality or change anything at the moment.

@gyfora Hi Gyula, very appreciate for your code review. And for the `is-bounded` comment, I answered some of my ideas, but I didn't resolve it.  Looking forward to the further discussion and review. Thx again :)

gyfora:
> > I had a few minor comments but it looks pretty good in general.
> > I also commented already on the is-bounded question, I think we should remove that logic from this PR as it does not introduce new functionality or change anything at the moment.
> 
> @gyfora Hi Gyula, very appreciate for your code review. And for the `is-bounded` comment, I answered some of my ideas, but I didn't resolve it. Looking forward to the further discussion and review. Thx again :)

As far as I understand the logic on the Flink side, the is-bounded flag does not affect the streaming execution and it should represent the actual source implementation.

The Kudu table source is always bounded right now. Therefore we should always return isBounded = true and not expose any options to change this as that might lead to incorrect/unexpected behavior.

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/table/KuduTableFactory.java
sebastianliu: already remove this unnecessary flag

sebastianliu:
@gyfora Hi Gyula, thx for your explanation. The is-bounded flag is really unnecessary and I have removed it. Hope you can help to review again. Thx for your time~

sebastianliu:
Hi @lresende , I would greatly appreciate if you could review this and all comments have been resolved so far. Let me know if you need further input from me.

PR #83, opened 2020-05-22
fix the bugs mentioned in bahir-231

zhoulii:
fix the bugs mentioned in [bahir-231](https://issues.apache.org/jira/browse/BAHIR-231)

### DIFF
--- a/flink-connector-activemq/src/test/java/org/apache/flink/streaming/connectors/activemq/ActiveMQConnectorITCase.java
+++ b/flink-connector-activemq/src/test/java/org/apache/flink/streaming/connectors/activemq/ActiveMQConnectorITCase.java
@@ -47,7 +47,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Random;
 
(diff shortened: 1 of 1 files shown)


lresende:
Is there a specific way to reproduce the fails? Just asking as the regular build tests are not failing.

zhoulii:
> Is there a specific way to reproduce the fails? Just asking as the regular build tests are not failing.

the test case did not set the prarllelism of flink task, so the default parallelism depends on the running enviroment. I run those test case on my pc, and the tasks are running with four threads by default, so when one thread is interrupted, the othe threads are still running, that is why method amqTopologyWithQueue and amqTopologyWithTopic will never end.

"the regular build tests are not failing": I think maybe the regular build tests run the flink job with single thread.  to reproduce the fails, you can set the parallelism of StreamExecutionEnvironment to  a value over 1. @lresende

zhoulii:
don't know why the test case amqTopologyWithCheckpointing could pass the regular build，it  actually failed when I run the test case. 

to reproduce this failure, you may run the test case on your own machine. @lresende

PR #88, opened 2020-07-20
Update KuduTableUtils.java

TheEros:
Support flink1.11sql read kudu


eskabetxe:
@gyfora could you check this

PR #89, opened 2020-07-26
BAHIR-240: replace docker test by testcontainer

eskabetxe:
replace docker scripts with the use of testcontainers


eskabetxe:
@lresende could you check

PR #98, opened 2020-08-27
[BAHIR-245] Add git .asf.yaml file

eskabetxe:
more info in [here](https://issues.apache.org/jira/browse/BAHIR-245)

@lresende could you check

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -257,6 +257,7 @@
           </executions>
           <configuration>
             <excludes>
+              <exclude>.asf.yaml</exclude>
               <exclude>.gitignore</exclude>
```
lresende, line 261: You shouldn't exclude this.

.asf.yaml
```diff
@@ -0,0 +1,10 @@
+github:
```
lresende, line 1: Yaml files allow asf license header using `#` at the beginning of the line.

eskabetxe: isnt this like '.gitignore' a configuration file...
this file wont be on released jars.

lresende: The configuration below on the pom disables license checking on files that do not allow license headers. all files that allow should have.

pom.xml
eskabetxe: done

.asf.yaml
eskabetxe: done

lresende (approved):

PR #100, opened 2020-11-22
[BAHIR-212] Add MQTT Connector for Flink

Ruguo:
Java library to integrate MQTT protocol and Apache Flink.
Main goal is to be able to read/write data from/to MQTT using the DataStream Flink's APIs.

### DIFF
--- a/flink-connector-mqtt/src/main/java/org/apache/flink/streaming/connectors/mqtt/MqttSource.java
+++ b/flink-connector-mqtt/src/main/java/org/apache/flink/streaming/connectors/mqtt/MqttSource.java
@@ -1,0 +1,229 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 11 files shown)


Ruguo:
@lresende @mbalassi @eskabetxe Appreciate for your time to review this PR. thanks!

flink-connector-mqtt/src/main/java/org/apache/flink/streaming/connectors/mqtt/MqttSource.java
```diff
@@ -0,0 +1,229 @@
...
+    @Override
+    public void open(Configuration parameters) throws Exception {
+        super.open(parameters);
+        if (connect()) {
+            this.isRunning = true;
+        }
+    }
+
+    private boolean connect() throws Exception {
+        this.serverUrl = properties.getProperty(MqttConfig.SERVER_URL);
```
eskabetxe, line 132: why we need to save this values?
count be used directly where needed?

flink-connector-mqtt/src/test/java/org/apache/flink/streaming/connectors/mqtt/MqttSinkTest.java
```diff
@@ -0,0 +1,106 @@
...
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.createLocalEnvironment();
+
+        Properties properties = new Properties();
+        properties.load(new FileInputStream(MqttSinkTest.class.getClassLoader().getResource(
+            "application.properties").getFile()));
+
+        Properties mqttProperties = new Properties();
+
+        // mqtt server url = tcp://<Org_ID>.messaging.internetofthings.ibmcloud.com:1883
+        mqttProperties.setProperty(MqttConfig.SERVER_URL,
```
eskabetxe, line 44: could we use some docker image instead of a external system?

eskabetxe (changes requested):
Thanks @yuruguo  for the contribution :)

I have added a few comments.

flink-connector-mqtt/src/main/java/org/apache/flink/streaming/connectors/mqtt/MqttSource.java
yuruguo: It's ok to use directly where needed, but this values may be used in multiple places, such as `properties.getProperty(MqttConfig.SERVER_URL)`, so saving them separately can ensure the code is concise.

PR #109, opened 2021-02-02
[BAHIR-260] add kudu table writer config. 

hackergin:
Add some  kudu write config that we can config flush interval, flush buffer size, ignore not found error, ignore-duplicate-error.

The default flush-interval is 1000ms.

default buffer-size is 1000,

default ignore-not-found error is false,

default ignore-duplicate is false.

### DIFF
--- a/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/connector/writer/KuduWriterConfig.java
+++ b/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/connector/writer/KuduWriterConfig.java
@@ -19,6 +19,7 @@ package org.apache.flink.connectors.kudu.connector.writer;
 import org.apache.flink.annotation.PublicEvolving;
 
 import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.kudu.client.AsyncKuduClient;
 
 import java.io.Serializable;
 import java.util.Objects;
@@ -102,7 +103,7 @@ public class KuduWriterConfig implements Serializable {
         private String masters;
         private FlushMode flushMode = FlushMode.AUTO_FLUSH_BACKGROUND;
         // Reference from AsyncKuduClientBuilder defaultOperationTimeoutMs.
-        private long timeout = 30000;
+        private long timeout = AsyncKuduClient.DEFAULT_OPERATION_TIMEOUT_MS;
         // Reference from AsyncKuduSession mutationBufferMaxOps 1000.
         private int maxBufferSize = 1000;
         // Reference from AsyncKuduSession flushIntervalMillis 1000.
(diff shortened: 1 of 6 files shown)


eskabetxe:
@hackergin thanks for your contribution
@gyfora could you check please

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/connector/writer/KuduWriterConfig.java
```diff
@@ -65,6 +101,11 @@ public String toString() {
     public static class Builder {
         private String masters;
         private FlushMode flushMode = FlushMode.AUTO_FLUSH_BACKGROUND;
+        private long timeout = 30000;
+        private int maxBufferSize = 1000;
+        private int flushInterval = 1000;
```
gyfora, line 106: Where did you take this defaults from? How will this affect the out of the box performance?

hackergin: These are from https://github.com/apache/kudu/blob/1eb2a2fbca/java/kudu-client/src/main/java/org/apache/kudu/client/AsyncKuduSession.java and  https://github.com/apache/kudu/blob/1eb2a2fbca/java/kudu-client/src/main/java/org/apache/kudu/client/AsyncKuduClient.java.  So, It will not affect the out of box performance.  I can add some  Annotation for them.

gyfora: sounds good

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/connector/writer/KuduWriterConfig.java
```diff
@@ -65,6 +101,16 @@ public String toString() {
     public static class Builder {
         private String masters;
         private FlushMode flushMode = FlushMode.AUTO_FLUSH_BACKGROUND;
+        // Reference from AsyncKuduClientBuilder defaultOperationTimeoutMs.
+        private long timeout = 30000;
```
eskabetxe, line 105: could we use AsyncKuduClientBuilder.DEFAULT_OPERATION_TIMEOUT_MS directly?

private long timeout = AsyncKuduClientBuilder.DEFAULT_OPERATION_TIMEOUT_MS;

hackergin: Sure,  I have update the code,  It seems that session-related parameters cannot be directly referenced.

eskabetxe:
@hackergin merged, thanks for you contribution

PR #113, opened 2021-03-10
Apache Pinot Connector Sink

mschroederi:
This MR implements an Apache Pinot connector sink using the new sink interface introduced in [FLIP-143](https://cwiki.apache.org/confluence/display/FLINK/FLIP-143%3A+Unified+Sink+API#FLIP143:UnifiedSinkAPI-SinkAPI).


flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotWriterSegment.java
```diff
@@ -0,0 +1,148 @@
...
+    public void write(IN element, long timestamp) throws IOException {
+        if (!this.acceptsElements()) {
+            throw new IllegalStateException("This PinotSegmentWriter does not accept any elements anymore.");
+        }
+        this.elements.add(element);
+        this.minTimestamp = Long.min(this.minTimestamp, timestamp);
+        this.maxTimestamp = Long.max(this.maxTimestamp, timestamp);
+
+        // Writes elements to local filesystem once the maximum number of items is reached
+        if (this.elements.size() == this.maxRowsPerSegment) {
```
eskabetxe, line 93: why we write to local?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/emulator/PinotEmulatorManager.java
```diff
@@ -0,0 +1,341 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.connectors.pinot.emulator;
+
+import com.github.dockerjava.api.DockerClient;
```
eskabetxe, line 21: could you use testcontainers?

eskabetxe (changes requested):
Thanks @mschroederi for the contribution :)

I have added a few comments.

flink-connector-pinot/pom.xml
```diff
@@ -0,0 +1,213 @@
...
+
+
+    <packaging>jar</packaging>
+
+    <!-- Allow users to pass custom connector versions -->
+    <properties>
+        <pinot.version>0.6.0</pinot.version>
+
+        <!-- Flink version -->
+        <flink.version>1.12.0</flink.version>
```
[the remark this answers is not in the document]
mschroederi: Totally right, this gets obsolete once #115 was merged

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,304 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(PinotControllerApi.class);
+    protected final String controllerHost;
+    protected final String controllerHostPort;
+
+    /**
+     * @param controllerHost Pinot controller's host
+     * @param controllerPort Pinot controller's port
+     */
+    public PinotControllerApi(String controllerHost, String controllerPort) {
+        this.controllerHost = checkNotNull(controllerHost);
```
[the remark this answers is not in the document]
mschroederi: You're right, thanks for noting

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/emulator/PinotEmulatorManager.java
mschroederi: Great advice, the manual testing setup was replaced by testcontainers

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,428 @@
...
+                LOG.info("Successfully created segment: {} in directory: {}", segmentName, indexDir);
+                if (_postCreationVerification) {
+                    LOG.info("Verifying the segment by loading it");
+                    ImmutableSegment segment = ImmutableSegmentLoader.load(indexDir, ReadMode.mmap);
+                    LOG.info("Successfully loaded segment: {} of size: {} bytes", segmentName,
+                            segment.getSegmentSizeBytes());
+                    segment.destroy();
+                }
+            } catch (Exception e) {
+                e.printStackTrace();
```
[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,428 @@
...
+                // Creates a segment with name `segmentName` in `segmentFile`
+                this.generateSegment(segmentData, segmentFile, true);
+
+                // Uploads the recently created segment to the Pinot controller
+                this.uploadSegment(segmentFile);
+
+                // Commit successful
+                return true;
+            } catch (IOException e) {
+                e.printStackTrace();
```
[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotWriterSegment.java
mschroederi: The current version of the `FileSystemAdapter` only supports copying local files to the shared FS and vice versa which should allow maximum flexibility when it comes to extensibility.
I nonetheless see your point that we can avoid unnecessary I/O here. Therefore, I'd suggest to change the `FileSystemAdapter`'s API to accept a list of serialized elements to write to the shared filesystem. Those adapters that need a local file to upload could first write to local disk before uploading and all others could directly write to the respective filesystem.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,144 @@
...
+     * @throws IOException
+     */
+    @Override
+    public void write(IN element, Context context) throws IOException {
+        final PinotWriterSegment<IN> inProgressSegment = this.getOrCreateInProgressSegment();
+        inProgressSegment.write(element, this.eventTimeExtractor.getEventTime(element, context));
+    }
+
+    /**
+     * Creates {@link PinotSinkCommittable}s from elements received via {@link #write}
```
fapaul, line 88: A bit more information what this method is doing in respect to the activeSegments would be great.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @param flush Flush all currently known elements into the {@link PinotSinkCommittable}s
+     * @return List of {@link PinotSinkCommittable} to process in {@link org.apache.flink.streaming.connectors.pinot.committer.PinotSinkGlobalCommitter}
+     * @throws IOException
+     */
+    @Override
+    public List<PinotSinkCommittable> prepareCommit(boolean flush) throws IOException {
```
fapaul, line 95: Again switch to debug logging.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Snapshots the current state to be stored within a checkpoint. As we do not need to save any
+     * information in snapshots, this method always returns an empty ArrayList.
+     *
+     * @return always an empty ArrayList
+     */
+    @Override
+    public List<Void> snapshotState() {
```
fapaul, line 134: What happens with the temporary written files?

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,144 @@
...
+        return new ArrayList<>();
+    }
+
+    /**
+     * Empty method, as we do not open any connections.
+     */
+    @Override
+    public void close() {
```
fapaul, line 142: Nit: clean up the temporary written files?

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,144 @@
...
+        if (latestSegment == null || !latestSegment.acceptsElements()) {
+            final PinotWriterSegment<IN> inProgressSegment = new PinotWriterSegment<>(this.maxRowsPerSegment, this.tempDirPrefix, this.jsonSerializer, this.fsAdapter);
+            this.activeSegments.add(inProgressSegment);
+            return inProgressSegment;
+        }
+        return latestSegment;
+    }
+
+    /**
+     * Snapshots the current state to be stored within a checkpoint. As we do not need to save any
```
fapaul, line 128:
```suggestion
     * As we do not need to save any
```

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotWriterSegment.java
fapaul: I think it is a good idea to directly write to the shared fs rather than having the local copy. Do you plan to implement it in this PR (extending the `FileSystemAdapater` interface should not be to complicated) or leave it as future optimization?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/EmulatedPinotSinkTest.java
```diff
@@ -0,0 +1,160 @@
...
+        env.setRuntimeMode(RuntimeExecutionMode.BATCH);
+        env.setParallelism(2);
+
+        List<SingleColumnTableRow> data = getTestData(12);
+        this.setupDataStream(env, data);
+
+        // Run
+        env.execute();
+
+        TimeUnit.SECONDS.sleep(5);
```
fapaul, line 62: I do not think you need a timeout for the batch mode because the job should go eventually into finished state.
You can do something along the lines of https://github.com/apache/flink/blob/bebf3b5a105dd4bc21882116570c6d71299269a6/flink-connectors/flink-connector-files/src/test/java/org/apache/flink/connector/file/sink/FileSinkITBase.java#L80

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/EmulatedPinotSinkTest.java
```diff
@@ -0,0 +1,160 @@
...
+        env.enableCheckpointing(50);
+
+        List<SingleColumnTableRow> data = getTestData(1000);
+        this.setupDataStream(env, data);
+
+        // Run
+        env.execute();
+
+        // Wait until the checkpoint was created and the segments were committed by the GlobalCommitter
+        TimeUnit.SECONDS.sleep(5);
```
fapaul, line 87: Maybe my suggestion for batch execution also works here because the streaming mode is finite when using the `NumberSequenceSource`

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/EmulatedPinotSinkTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * E2e tests for Pinot Sink using BATCH and STREAMING execution mode
+ */
+public class EmulatedPinotSinkTest extends PinotTestBase {
```
fapaul, line 43: Overall the timeouts make the code unreliable and make is runtime somewhat unpredictable and flaky on slower systems. If possible you should replace all with better synchronization mechanics

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+import java.io.IOException;
+import java.io.Serializable;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Base class for PinotSink e2e tests
+ */
+@Testcontainers
+public class PinotTestBase extends TestLogger implements Serializable {
```
fapaul, line 48: Why Serializable?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/EmulatedPinotSinkTest.java
fapaul, line 43:
```suggestion
public class PinotSinkTest extends PinotTestBase {
```

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+
+    public static final String DOCKER_IMAGE_NAME = "apachepinot/pinot:0.6.0";
+    public static final Integer PINOT_INTERNAL_BROKER_PORT = 8000;
+    public static final Integer PINOT_INTERNAL_CONTROLLER_PORT = 9000;
```
fapaul, line 52: private?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Creates the Pinot testcontainer. We delay the start of tests until Pinot has started all
+     * internal components. This is identified through a log statement.
+     */
+    @Container
+    public GenericContainer<?> pinot = new GenericContainer<>(DockerImageName.parse(DOCKER_IMAGE_NAME))
+            .withCommand("QuickStart", "-type", "batch")
+            .withExposedPorts(PINOT_INTERNAL_BROKER_PORT, PINOT_INTERNAL_CONTROLLER_PORT)
+            .waitingFor(Wait
+                    .forLogMessage(".*You can always go to http://localhost:9000 to play around in the query console.*\\n", 1)
```
fapaul, line 68: Nit: Maybe wait until the port is consumable? When pinot is updated in the future this check might be hard to fix.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+    public void tearDown() throws Exception {
+        pinotHelper.deleteTable(TABLE_CONFIG, TABLE_SCHEMA);
+    }
+
+    /**
+     * Returns the host the Pinot container is available at
+     *
+     * @return Pinot container host
+     */
+    public String getPinotHost() {
```
fapaul, line 100: protected?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+        return this.pinot.getHost();
+    }
+
+
+    /**
+     * Returns the Pinot controller port from the container ports.
+     *
+     * @return Pinot controller port
+     */
+    public String getPinotControllerPort() {
```
fapaul, line 110: protected?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+        return this.pinot.getMappedPort(PINOT_INTERNAL_CONTROLLER_PORT).toString();
+    }
+
+    /**
+     * Returns the Pinot broker port from the container ports.
+     *
+     * @return Pinot broker port
+     */
+    public String getPinotBrokerPort() {
```
fapaul, line 119: protected?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,244 @@
...
+        @Override
+        public String toJson(SingleColumnTableRow element) {
+            return JsonUtils.objectToJsonNode(element).toString();
+        }
+    }
+
+    /**
+     * Pinot table configuration helpers.
+     */
+    static class PinotTableConfig {
```
fapaul, line 193: private?

fapaul (changes requested):
@mschroederi overall great effort! In summary I found a few overarching things which need to be addressed:

- Logging: For frequent logging statement use the DEBUG level and only use INFO for during initializations or similarly rare activities. ERROR on error...
- Please only use `this` if needed currently it is all over the codebase :)
- Change the `PinotControllerApi` to only be instantiated once with the underlying http client
- Revisit the visibility of your classes. In the end a user only cares about the Sink and does not want to see all other classes (package-private, private).

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+     * Issues a request to the Pinot controller API.
+     *
+     * @param request Request to issue
+     * @return Api response
+     * @throws IOException
+     */
+    private ApiResponse execute(HttpRequestBase request) throws IOException {
+        ApiResponse result;
+
+        try (CloseableHttpClient httpClient = HttpClients.createDefault();
```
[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+                File indexDir = new File(outDir, segmentName);
+                LOG.info("Successfully created segment: {} in directory: {}", segmentName, indexDir);
+                if (_postCreationVerification) {
+                    LOG.info("Verifying the segment by loading it");
+                    ImmutableSegment segment = ImmutableSegmentLoader.load(indexDir, ReadMode.mmap);
+                    LOG.info("Successfully loaded segment: {} of size: {} bytes", segmentName,
+                            segment.getSegmentSizeBytes());
+                    segment.destroy();
+                }
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
mschroederi: The `SegmentIndexCreationDriverImpl` (`init` and `build`) and the `ImmutableSegmentLoader` (`load`) both throw generic Exceptions, so we have to catch those generic Exceptions here.
I've added a comment to explain this

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+     *
+     * @param path Path to POST to
+     * @param body Request's body
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse post(String path, String body) throws IOException {
+        HttpPost httppost = new HttpPost(this.controllerHostPort + path);
+        httppost.setEntity(new StringEntity(body, ContentType.APPLICATION_JSON));
+        LOG.info("Posting string entity {} to {}", body, path);
```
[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+    /**
+     * Issues a GET request to the Pinot controller API.
+     *
+     * @param path Path to GET from
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse get(String path) throws IOException {
+        HttpGet httpget = new HttpGet(this.controllerHostPort + path);
+        LOG.info("Sending GET request to {}", path);
```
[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+    /**
+     * Issues a DELETE request to the Pinot controller API.
+     *
+     * @param path Path to issue DELETE request to
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse delete(String path) throws IOException {
+        HttpDelete httpdelete = new HttpDelete(this.controllerHostPort + path);
+        LOG.info("Sending DELETE request to {}", path);
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+        if (res.statusLine.getStatusCode() != 200) {
+            throw new PinotControllerApiException(res.responseBody);
+        }
+
+        try {
+            schema = JsonUtils.stringToObject(res.responseBody, Schema.class);
+        } catch (Exception e) {
+            throw new IllegalStateException("Caught exception while reading schema from Pinot Controller's response: " + res.responseBody, e);
+        }
+        LOG.info("Retrieved schema: {}", schema.toSingleLineJsonString());
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+     * Fetches a Pinot table's configuration via the Pinot controller API.
+     *
+     * @param tableName Target table's name
+     * @return Pinot table configuration
+     * @throws IOException
+     */
+    public TableConfig getTableConfig(String tableName) throws IOException {
+        TableConfig tableConfig;
+        ApiResponse res = this.get(String.format("/tables/%s", tableName));
+        LOG.info("Get table config request for table {} returned {}", tableName, res.responseBody);
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+
+        try {
+            String tableConfigAsJson = JsonUtils.stringToJsonNode(res.responseBody).get("OFFLINE").toString();
+            tableConfig = JsonUtils.stringToObject(tableConfigAsJson, TableConfig.class);
+        } catch (Exception e) {
+            throw new IllegalStateException("Caught exception while reading table config from Pinot Controller's response: " + res.responseBody, e);
+        }
+        LOG.info("Retrieved table config: {}", tableConfig.toJsonString());
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+
+        // Extract all data file paths and the overall minimum and maximum timestamps
+        // from all committables
+        for (PinotSinkCommittable committable : committables) {
+            dataFilePaths.add(committable.getDataFilePath());
+            minTimestamp = Long.min(minTimestamp, committable.getMinTimestamp());
+            maxTimestamp = Long.max(maxTimestamp, committable.getMaxTimestamp());
+        }
+
+        LOG.info("Combined {} committables into one global committable", committables.size());
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+         * Downloads a segment from the shared file system via {@code fsAdapter}, generates a segment
+         * and finally uploads the segment to the Pinot controller
+         *
+         * @return True if the commit succeeded
+         */
+        @Override
+        public Boolean call() {
+            try {
+                // Download data file from the shared filesystem
+                LOG.info("Downloading data file {} from shared file system...", dataFilePath);
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+            segmentGeneratorConfig.setFormat(FileFormat.JSON);
+            segmentGeneratorConfig.setOutDir(outDir.getPath());
+            segmentGeneratorConfig.setTableName(tableConfig.getTableName());
+
+            try {
+                SegmentIndexCreationDriver driver = new SegmentIndexCreationDriverImpl();
+                driver.init(segmentGeneratorConfig);
+                driver.build();
+                File indexDir = new File(outDir, segmentName);
+                LOG.info("Successfully created segment: {} in directory: {}", segmentName, indexDir);
```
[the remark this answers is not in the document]
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/segment/name/SimpleSegmentNameGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+ * </ul>
+ */
+public class SimpleSegmentNameGenerator extends PinotSinkSegmentNameGenerator {
+
+    private final String tableName;
+    private final String segmentNamePostfix;
+
+    public SimpleSegmentNameGenerator(String tableName, String segmentNamePostfix) {
+        this.tableName = tableName;
+        this.segmentNamePostfix = segmentNamePostfix;
```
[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/serializer/PinotSinkCommittableSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+        return CURRENT_VERSION;
+    }
+
+    @Override
+    public byte[] serialize(PinotSinkCommittable pinotSinkCommittable) {
+        return SerializationUtils.serialize(pinotSinkCommittable);
+    }
+
+    @Override
+    public PinotSinkCommittable deserialize(int i, byte[] bytes) {
```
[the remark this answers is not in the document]
mschroederi: updated to use custom serializer & deserializer with version checking

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/serializer/PinotSinkGlobalCommittableSerializer.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.flink.streaming.connectors.pinot.serializer;
+
+import org.apache.commons.lang3.SerializationUtils;
+import org.apache.flink.core.io.SimpleVersionedSerializer;
+import org.apache.flink.streaming.connectors.pinot.committer.PinotSinkGlobalCommittable;
+
+/**
+ * Serializer for {@link PinotSinkGlobalCommittable}
+ */
+public class PinotSinkGlobalCommittableSerializer implements SimpleVersionedSerializer<PinotSinkGlobalCommittable> {
```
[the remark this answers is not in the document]
mschroederi: updated to use custom serializer & deserializer with version checking

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
mschroederi: now using debug instead of info all over the place

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
mschroederi: really good question -> removed

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+    /**
+     * Identifies global committables that need to be re-committed from a list of recovered committables.
+     *
+     * @param globalCommittables List of global committables that are checked for required re-commit
+     * @return List of global committable that need to be re-committed
+     * @throws IOException
+     */
+    @Override
+    public List<PinotSinkGlobalCommittable> filterRecoveredCommittables(List<PinotSinkGlobalCommittable> globalCommittables) throws IOException {
+        PinotControllerApi controllerApi = new PinotControllerApi(this.pinotControllerHost, this.pinotControllerPort);
```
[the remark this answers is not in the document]
mschroederi: see above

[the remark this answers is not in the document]
mschroederi: done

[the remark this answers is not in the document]
mschroederi: done

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+                controllerApi.deleteSegment(tableName, existingSegment);
+            }
+
+            // We use a thread pool in order to parallelize the segment creation and segment upload
+            ExecutorService pool = Executors.newCachedThreadPool();
+            Set<Future<Boolean>> resultFutures = new HashSet<>();
+
+            // Commit all segments in globalCommittable
+            int sequenceId = 0;
+            for (String dataFilePath : globalCommittable.getDataFilePaths()) {
```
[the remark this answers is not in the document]
mschroederi: good hint. Thanks 👍

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/segment/name/PinotSinkSegmentNameGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.pinot.core.segment.name.SegmentNameGenerator;
+
+import javax.annotation.Nullable;
+import java.io.Serializable;
+
+/**
+ * Defines the segment name generator interface that is used to generate segment names.
+ */
+public abstract class PinotSinkSegmentNameGenerator implements SegmentNameGenerator, Serializable {
```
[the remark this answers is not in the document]
mschroederi: We require the segment name generator to be serializable. This class just defines the required interface which users can inherit from to create their custom name generator.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+                    // In case any of the segment commits wasn't successful we mark the whole
+                    // globalCommittable as failed
+                    if (!wasSuccessful.get()) {
+                        failedCommits.add(globalCommittable);
+                        // Once any of the commits failed, we do not need to check the remaining
+                        // ones, as we try to commit the globalCommittable next time
+                        break;
+                    }
+                }
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
mschroederi: `InterruptedException | ExecutionException` possibly thrown when trying to resolve the `Future`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+                // Submits the segment committer to the thread pool
+                resultFutures.add(pool.submit(segmentCommitter));
+            }
+
+            try {
+                for (Future<Boolean> wasSuccessful : resultFutures) {
+                    // In case any of the segment commits wasn't successful we mark the whole
+                    // globalCommittable as failed
+                    if (!wasSuccessful.get()) {
+                        failedCommits.add(globalCommittable);
```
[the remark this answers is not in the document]
mschroederi: The idea of returning the failed commits here, is that the "commit-with-retry" pattern can be used once implemented. I would suppose that the number of retries will be definable by the user to prevent endless retrying commits.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
mschroederi: The `PinotWriterSegment` now directly writes to the shared filesystem via `FileSystemAdapter`, so there aren't any temporary local files used anymore in the SinkWriter.
Files stored on the shared filesystem will be deleted once the commit was successful.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
mschroederi: No temp files anymore in the SinkWriter

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotWriterSegment.java
mschroederi: This is now implemented in the current version of this PR. The `FileSystemAdapter` is now capable of writing and reading directly from the shared filesystem. Moreover, it supports deleting files once the committables were successfully committed.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+    public TableConfig getTableConfig(String tableName) throws IOException {
+        TableConfig tableConfig;
+        ApiResponse res = this.get(String.format("/tables/%s", tableName));
+        LOG.info("Get table config request for table {} returned {}", tableName, res.responseBody);
+
+        try {
+            String tableConfigAsJson = JsonUtils.stringToJsonNode(res.responseBody).get("OFFLINE").toString();
+            tableConfig = JsonUtils.stringToObject(tableConfigAsJson, TableConfig.class);
+        } catch (Exception e) {
+            throw new IllegalStateException("Caught exception while reading table config from Pinot Controller's response: " + res.responseBody, e);
```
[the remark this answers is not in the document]
mschroederi: You're right, I've replaced `IllegalStateException` with `PinotControllerApiException`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,218 @@
...
+
+    /**
+     * Issues a POST request to the Pinot controller API.
+     *
+     * @param path Path to POST to
+     * @param body Request's body
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse post(String path, String body) throws IOException {
```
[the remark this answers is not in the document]
mschroederi: Great hint! We now have a `PinotControllerClient` and `PinotControllerHttpClient`. The later provides an external API, e.g. used by the `PinotControllerClient`, to send http requests to the Pinot controller. The `PinotControllerClient` provides high-level functionality.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+         */
+        @Override
+        public Boolean call() {
+            try {
+                // Download data file from the shared filesystem
+                LOG.info("Downloading data file {} from shared file system...", dataFilePath);
+                File segmentData = fsAdapter.copyToLocalFile(dataFilePath);
+                LOG.info("Successfully downloaded data file {} from shared file system", dataFilePath);
+
+                File segmentFile = Files.createTempDirectory(this.tempDirPrefix).toFile();
```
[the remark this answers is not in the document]
mschroederi: The global committer now creates one temp directory which is deleted up onClose. The temp files created during segment commit are deleted after successful or failed commits.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+     *
+     * @param globalCommittables List of global committables
+     * @return Global committables whose commit failed
+     * @throws IOException
+     */
+    @Override
+    public List<PinotSinkGlobalCommittable> commit(List<PinotSinkGlobalCommittable> globalCommittables) throws IOException {
+        // Retrieve the Pinot table schema and the Pinot table config from the Pinot controller
+        PinotControllerApi controllerApi = new PinotControllerApi(this.pinotControllerHost, this.pinotControllerPort);
+        Schema tableSchema = controllerApi.getSchema(this.tableName);
```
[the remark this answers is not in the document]
mschroederi: I guess supporting changing schemas over time is out-of-scope for this PR, as it would require versioning of tuples etc.
The schema and tableConfig retrieval is now handled in the constructor.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+            CommitStatus commitStatus = this.getCommitStatus(globalCommittable);
+            for (String existingSegment : commitStatus.getExistingSegmentNames()) {
+                // Some but not all segments were already committed. As we cannot assure the data
+                // files containing the same data as originally when recovering from failure,
+                // we delete the already committed segments in order to recommit them later on.
+                controllerApi.deleteSegment(tableName, existingSegment);
+            }
+
+            // We use a thread pool in order to parallelize the segment creation and segment upload
+            ExecutorService pool = Executors.newCachedThreadPool();
```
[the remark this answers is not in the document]
mschroederi: I totally agree that it would have been much cleaner if the `SinkGlobalCommitter` was able to notify the `SinkWriter`. This would e.g. make the external filesystem obsolete.  
But this dataflow is currently not supported by the Flink sink interface.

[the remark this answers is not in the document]
mschroederi: Nonetheless, I guess the threading model makes the `PinotSink` much more usable. I've started some benchmarks and saw a huge increase in throughput while keeping the event time latency stable in comparison to the single threaded `PinotSinkGlobalCommitter`.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,426 @@
...
+
+        for (PinotSinkGlobalCommittable globalCommittable : globalCommittables) {
+            // Make sure to remove all previously committed segments in globalCommittable
+            // when recovering from failure
+            CommitStatus commitStatus = this.getCommitStatus(globalCommittable);
+            for (String existingSegment : commitStatus.getExistingSegmentNames()) {
+                // Some but not all segments were already committed. As we cannot assure the data
+                // files containing the same data as originally when recovering from failure,
+                // we delete the already committed segments in order to recommit them later on.
+                controllerApi.deleteSegment(tableName, existingSegment);
```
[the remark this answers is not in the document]
mschroederi: Totally right, `filterRecoveredCommittables` already handles the recovery and identified global committables that need to be re-committed, including the deleting of existing segment artifacts.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerApi.java
```diff
@@ -0,0 +1,304 @@
...
+     * @throws IOException
+     */
+    private ApiResponse execute(HttpRequestBase request) throws IOException {
+        ApiResponse result;
+
+        try (CloseableHttpClient httpClient = HttpClients.createDefault();
+             CloseableHttpResponse response = httpClient.execute(request)) {
+
+
+            String body = EntityUtils.toString(response.getEntity());
```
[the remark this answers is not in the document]
mschroederi: `httpClient.execute(request)` will raise an `HttpHostConnectException`(`IOException`), if the connection could not be established.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/EmulatedPinotSinkTest.java
mschroederi: The issue here is that Pinot might take some time in order to index and distribute the recently uploaded segments (independent from Streaming or Batching mode). 
In order to remove complexity from the actual testing code, I've now implemented a `checkForDataInPinotWithRetry` method which uses a linear retry backoff delay to regularly call `checkForDataInPinot` up to a given timeout.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
mschroederi: We now make use of `HttpWaitStrategy` which checks via the public Pinot controller API whether all required instances in Pinot (Controller, Broker and Server) have been successfully startet up.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * As we do not need to save any information in snapshots,
+     * this method always returns an empty ArrayList.
+     *
+     * @return always an empty ArrayList
+     */
+    @Override
+    public List<Void> snapshotState() {
```
fapaul, line 143: I would expect that we have to save the path to the active files on every checkpoint.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestHelper.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * @param host           Host the Pinot controller and broker are accessible at
+     * @param controllerPort The Pinot controller's external port at {@code host}
+     * @param brokerPort     A Pinot broker's external port at {@code host}
+     */
+    public PinotTestHelper(String host, String controllerPort, String brokerPort) {
+        this.host = host;
+        this.brokerPort = brokerPort;
+        httpClient = new PinotControllerHttpClient(host, controllerPort);
```
fapaul, line 50:
```suggestion
        this.httpClient = new PinotControllerHttpClient(host, controllerPort);
```

fapaul (changes requested):
Thanks addressing all the comments! Overall it looks already much better. I left a few minor comments and one question regarding the checkpointing in the SinkWriter.

flink-connector-pinot/pom.xml
```diff
@@ -0,0 +1,225 @@
...
+                        <configuration>
+                            <transformers>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                                    <resource>reference.conf</resource>
+                                </transformer>
+                            </transformers>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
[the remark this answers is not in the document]
mschroederi: The [main pom.xml](https://github.com/apache/bahir-flink/blob/b618a77cca2e44be86f18607e21a6b2d2bf8dd39/pom.xml#L237) excludes the `org.codehaus.groovy` dependency, but `org.apache.pinot:pinot-tools` requires it, so we have to override the exclude.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerHttpClient.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Issues a DELETE request to the Pinot controller API.
+     *
+     * @param path Path to issue DELETE request to
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse delete(String path) throws IOException {
```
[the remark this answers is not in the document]
mschroederi: We need to expose this externally, used e.g. in `PinotControllerClient`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerHttpClient.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Issues a POST request to the Pinot controller API.
+     *
+     * @param path Path to POST to
+     * @param body Request's body
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse post(String path, String body) throws IOException {
```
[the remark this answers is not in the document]
mschroederi: We need to expose this externally, used e.g. in `PinotControllerClient`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerHttpClient.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * Issues a GET request to the Pinot controller API.
+     *
+     * @param path Path to GET from
+     * @return API response
+     * @throws IOException
+     */
+    protected ApiResponse get(String path) throws IOException {
```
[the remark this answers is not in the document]
mschroederi: We need to expose this externally, used e.g. in `PinotControllerClient`

flink-connector-pinot/pom.xml
```diff
@@ -0,0 +1,225 @@
...
+    <name>flink-connector-pinot</name>
+
+
+    <packaging>jar</packaging>
+
+    <!-- Allow users to pass custom connector versions -->
+    <properties>
+        <pinot.version>0.6.0</pinot.version>
+
+        <testcontainers.version>1.15.2</testcontainers.version>
```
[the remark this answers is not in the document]
mschroederi: You're right. Thanks for noting 👍

flink-connector-pinot/README.md
```diff
@@ -0,0 +1,126 @@
...
+| `tableName`            | Target Pinot table's name                                                        |
+| `maxRowsPerSegment`    | Maximum number of rows to be stored within a Pinot segment                       |
+| `tempDirPrefix`         | Prefix for temp directories used                                                  |
+| `jsonSerializer`       | Serializer used to convert elements to JSON                                      |
+| `eventTimeExtractor`   | Defines the way event times are extracted from received objects                   |
+| `segmentNameGenerator` | Pinot segment name generator                                                     |
+| `fsAdapter`            | Filesystem adapter used to save files for sharing files across nodes               |
+| `numCommitThreads`     | Number of threads used in the `PinotSinkGlobalCommitter` for committing segments |
+
+## Architecture
```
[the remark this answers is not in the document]
mschroederi: Totally makes sense 👍

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
mschroederi: Not sure why we would need to store anything here. Could you elaborate on what you consider to be 'active files' ?
From my understanding we cannot reuse previously written `PinotWriterSegment`s, as elements might get sent to different subTasks when recovering from failure.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotControllerHttpClient.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @param controllerHost Pinot controller's host
+     * @param controllerPort Pinot controller's port
+     */
+    public PinotControllerHttpClient(String controllerHost, String controllerPort) {
+        checkNotNull(controllerHost);
+        checkNotNull(controllerPort);
+        controllerHostPort = String.format("http://%s:%s", controllerHost, controllerPort);
+        httpClient = HttpClients.createDefault();
```
fapaul, line 55: `this.`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotSink.java
```diff
@@ -0,0 +1,361 @@
...
+ * Please note that the maximum segment size that can be handled by this sink is limited by the
+ * lower bound of memory available at each subTask.
+ * Each subTask holds a list of {@link PinotWriterSegment}s of which at most one is active. An
+ * active {@link PinotWriterSegment} is capable of accepting at least one more element. If a
+ * {@link PinotWriterSegment} switches from active to inactive it flushes its
+ * {@code maxRowsPerSegment} elements to disk. The data file is stored in the local filesystem's
+ * temporary directory and contains serialized elements. We use the {@link JsonSerializer} to
+ * serialize elements to JSON.
+ *
+ * <p>On checkpointing all inactive {@link PinotWriterSegment} are transformed into committables. As
```
fapaul, line 73: You also have to deal with the active segments :) Once this is implemented update the doc string

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/PinotSink.java
```diff
@@ -0,0 +1,361 @@
...
+        String pinotControllerHost;
+        String pinotControllerPort;
+        String tableName;
+        int maxRowsPerSegment;
+        String tempDirPrefix = "flink-connector-pinot";
+        JsonSerializer<IN> jsonSerializer;
+        EventTimeExtractor<IN> eventTimeExtractor;
+        SegmentNameGenerator segmentNameGenerator;
+        FileSystemAdapter fsAdapter;
+        int numCommitThreads = 4;
```
fapaul, line 236: Nit: Use a constant variable `static final ...` for the magic number `4` and call it something like `DEFAULT_COMMIT_THREADS`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,469 @@
...
+        // Create directory that temporary files will be stored in
+        tempDirectory = Files.createTempDirectory(tempDirPrefix).toFile();
+
+        // Retrieve the Pinot table schema and the Pinot table config from the Pinot controller
+        tableSchema = pinotControllerClient.getSchema(tableName);
+        tableConfig = pinotControllerClient.getTableConfig(tableName);
+
+        // We use a thread pool in order to parallelize the segment creation and segment upload
+        checkArgument(numCommitThreads > 0);
+        pool = Executors.newFixedThreadPool(numCommitThreads);
```
fapaul, line 110: `this.`

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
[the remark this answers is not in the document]
fapaul: Ah, I forgot about it...

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/committer/PinotSinkGlobalCommitter.java
```diff
@@ -0,0 +1,469 @@
...
+                if (_postCreationVerification) {
+                    LOG.debug("Verifying the segment by loading it");
+                    ImmutableSegment segment = ImmutableSegmentLoader.load(indexDir, ReadMode.mmap);
+                    LOG.debug("Successfully loaded segment: {} of size: {} bytes", segmentName,
+                            segment.getSegmentSizeBytes());
+                    segment.destroy();
+                }
+            } catch (Exception e) {
+                // SegmentIndexCreationDriverImpl throws generic Exceptions during init and build
+                // ImmutableSegmentLoader throws generic Exception during load
```
fapaul, line 441: Nit: move this comment above the exception.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/external/EventTimeExtractor.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.Serializable;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Defines the interface for event time extractors
+ *
+ * @param <IN> Type of incoming elements
+ */
+public abstract class EventTimeExtractor<IN> implements Serializable {
```
fapaul, line 31: interface?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
```diff
@@ -0,0 +1,194 @@
...
+        List<SingleColumnTableRow> data = getTestData(1000);
+        this.setupDataStream(env, data);
+
+        // Run
+        env.execute();
+
+        // We only expect the first 100 elements to be already committed to Pinot.
+        // The remaining would follow once we increase the input data size.
+        // The stream executions stops once the last input tuple was sent to the sink.
+        checkForDataInPinotWithRetry(data, 100, 20);
```
fapaul, line 87: Why do we not get all records? Since we use `env.execute()` which blocks until the job finishes I would expect all records in pinot.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
```diff
@@ -0,0 +1,194 @@
...
+        env.setRuntimeMode(RuntimeExecutionMode.BATCH);
+        env.setParallelism(2);
+
+        List<SingleColumnTableRow> data = getTestData(12);
+        this.setupDataStream(env, data);
+
+        // Run
+        env.execute();
+
+        checkForDataInPinotWithRetry(data, data.size(), 20);
```
fapaul, line 62: Probably the same question again why not all records are present.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,262 @@
...
+    protected String getPinotControllerPort() {
+        return this.pinot.getMappedPort(PINOT_INTERNAL_CONTROLLER_PORT).toString();
+    }
+
+    /**
+     * Returns the Pinot broker port from the container ports.
+     *
+     * @return Pinot broker port
+     */
+    protected String getPinotBrokerPort() {
```
fapaul, line 137: private?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,262 @@
...
+            this._timestamp = timestamp;
+        }
+    }
+
+
+    /**
+     * EventTimeExtractor for {@link SingleColumnTableRow} used in e2e tests.
+     * Extracts the timestamp column from {@link SingleColumnTableRow}.
+     */
+    protected static class SingleColumnTableRowEventTimeExtractor extends EventTimeExtractor<SingleColumnTableRow> {
```
fapaul, line 179: Move to `PinotSinkTest` because it is only used there?

fapaul (commented):
It probably requires another round of nitty cleanups and fixing the checkpointing of the writer but the rest seems to be good 👍

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
mschroederi: As the `SingleColumnTableRowEventTimeExtractor` will only be used within sink tests it makes sense to move it there.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
mschroederi: You are totally right, many thanks for pointing that out! I've included a `PinotWriterState` that snapshots the necessary information (elements, min&max timestamps) from the `PinotWriterSegment` which is currently in-progress.  
Upon `PinotSinkWriter` creation the state would then get recovered and transformed into a new in-progress `PinotWriterSegment`.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
mschroederi: I just saw that I missed the comment on the `MiniCluster` execution from the last review cycle. After finally switching to that execution form, everything works fine - including that the jobs are only stopped once all elements were committed.

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotSinkWriter.java
```diff
@@ -0,0 +1,176 @@
...
+     * committed) elements can be recovered later on in case of a failure.
+     *
+     * @return A list containing at most one PinotSinkWriterState
+     */
+    @Override
+    public List<PinotSinkWriterState> snapshotState() {
+        final PinotWriterSegment<IN> latestSegment = Iterables.getLast(activeSegments, null);
+        if (latestSegment == null || !latestSegment.acceptsElements()) {
+            return new ArrayList<>();
+        }
```
fapaul, line 150: Looks solid 👍

flink-connector-pinot/src/main/java/org/apache/flink/streaming/connectors/pinot/writer/PinotWriterSegment.java
```diff
@@ -0,0 +1,153 @@
...
+    private String dataPathOnSharedFS;
+    private long minTimestamp = Long.MAX_VALUE;
+    private long maxTimestamp = Long.MIN_VALUE;
+
+    /**
+     * @param maxRowsPerSegment Maximum number of rows to be stored within a Pinot segment
+     * @param jsonSerializer    Serializer used to convert elements to JSON
+     * @param fsAdapter         Filesystem adapter used to save files for sharing files across nodes
+     */
+    protected PinotWriterSegment(int maxRowsPerSegment, JsonSerializer<IN> jsonSerializer, FileSystemAdapter fsAdapter) {
```
fapaul, line 58:
```suggestion
   PinotWriterSegment(int maxRowsPerSegment, JsonSerializer<IN> jsonSerializer, FileSystemAdapter fsAdapter) {
```
package-private?

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Executes a given JobGraph on a MiniCluster.
+     *
+     * @param jobGraph JobGraph to execute
+     * @throws Exception
+     */
+    private void executeOnMiniCluster(JobGraph jobGraph) throws Exception {
+        final Configuration config = new Configuration();
+        config.setString(RestOptions.BIND_PORT, "18081-19000");
+        final MiniClusterConfiguration cfg =
```
fapaul, line 117: I think it makes more sense to specify the mini cluster with the help of `MiniClusterWithClientResource` and make it a `@ClassRule`. This way the cluster is only instantiated once and it drastically speeds up the tests. https://github.com/apache/flink/blob/fd8bae505281c443d4a74b97a6666845744921e7/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java#L95

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
```diff
@@ -0,0 +1,232 @@
...
+    private void checkForDataInPinotWithRetry(List<SingleColumnTableRow> data, int retryTimeoutInSeconds) throws InterruptedException {
+        long endTime = System.currentTimeMillis() + 1000L * retryTimeoutInSeconds;
+        // Use max 10 retries with linear retry backoff delay
+        long retryDelay = 1000L / 10 * retryTimeoutInSeconds;
+        do {
+            try {
+                checkForDataInPinot(data);
+                // In case of no error, we can skip further retries
+                return;
+            } catch (AssertionFailedError | PinotControllerApiException e) {
```
fapaul, line 181: Do you still allow the `AssertionFailedError` I would assume once the job is finished all data is in pinot.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,236 @@
...
+    protected static final String TABLE_NAME = TABLE_CONFIG.getTableName();
+    protected static final Schema TABLE_SCHEMA = PinotTableConfig.getTableSchema();
+    protected static PinotTestHelper pinotHelper;
+
+    /**
+     * Creates the Pinot testcontainer. We delay the start of tests until Pinot has started all
+     * internal components. This is identified through a log statement.
+     */
+    @Container
+    public GenericContainer<?> pinot = new GenericContainer<>(DockerImageName.parse(DOCKER_IMAGE_NAME))
```
fapaul, line 65: You should try to make this variable `static` then it is shared between tests otherwise every test starts a new pinot container.

fapaul (commented):
The checkpointing logic in the writer looks solid now, good job. I left some further comments regarding the test setup, to reduce the overall runtime.
Once this is fixed and the offline discussed failure test is done I think this PR is good to go.

fapaul:
@mschroederi can you please close the fixed discussions? ;)

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
mschroederi: Great hint!

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
mschroederi: It could still be that Pinot just hasn't (fully) finished indexing all the pushed segments when we check for the data.
Nonetheless, I spotted an error here: It could be that we catch the `AssertionFailedError`, sleep for some time and then finally do not execute the check once again as the `retryTimeoutInSeconds` was exceeded. So although the test actually should have failed (data not in Pinot), we would not raise an exception.

In order to prevent this, I've added a final check whether all data is in Pinot.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -85,6 +92,35 @@
                             .withStartupTimeout(Duration.ofSeconds(180))
             );
 
+    @ClassRule
+    public static final MiniCluster MINI_CLUSTER = new MiniCluster(
```
fapaul, line 96: Either remove the `@ClassRule` or, my preference, use https://github.com/apache/flink/blob/master/flink-test-utils-parent/flink-test-utils/src/main/java/org/apache/flink/test/util/MiniClusterWithClientResource.java. Currently the `@ClassRule` has no effect because the `MiniCluster` does not implement the required interfaces therefore you have to call `@BeforeAll` and `@AfterAll` manually.

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
```diff
@@ -0,0 +1,396 @@
...
+     */
+    private DataStream<SingleColumnTableRow> setupFailingMapper(DataStream<SingleColumnTableRow> dataStream, int failOnceAtNthElement) {
+        AtomicInteger messageCounter = new AtomicInteger(0);
+
+        return dataStream.map(element -> {
+            if (!hasFailedOnce.get() && messageCounter.incrementAndGet() == failOnceAtNthElement) {
+                hasFailedOnce.set(true);
+                // Wait more than STREAMING_CHECKPOINTING_INTERVAL to ensure
+                // that at least one checkpoint was created before raising the exception.
+                Thread.sleep(4 * STREAMING_CHECKPOINTING_INTERVAL);
```
fapaul, line 231: Maybe you can find a way to also remove this timeout :) One idea would be to make your map function implement `CheckpointedFunction` which allows you to receive the checkpoint notification.

I also do not really know why it is important for the failure to have at-least one checkpoint.

fapaul (approved):
Thanks for all your efforts, you did a great job!

Besides my final comment, this PR looks ready to merge from my side. Now we only need to wait for the final approval of Bahir committer or PMC.

Please squash all your changes in a single commit, to make the committer's life easier ;)

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotSinkTest.java
mschroederi: I've removed the timeout by extending the source to support exception raising and added the following comment to explain why we wait for a snapshot creation. I hope this clears things up.
> This allows to check whether the snapshot creation and failure recovery in
> {@link org.apache.flink.streaming.connectors.pinot.writer.PinotSinkWriter} works properly,
> respecting the already committed elements and those that are stored in an active
> {@link org.apache.flink.streaming.connectors.pinot.writer.PinotWriterSegment}. Committed 
> elements must not be saved to the snapshot while those in an active segment must be saved 
> to the snapshot in order to enable later-on recovery.

mschroederi:
> Thanks for all your efforts, you did a great job!
> 
> Besides my final comment, this PR looks ready to merge from my side. Now we only need to wait for the final approval of Bahir committer or PMC.
> 
> Please squash all your changes in a single commit, to make the committer's life easier ;)

@fapaul Thanks for your support during this whole process!

fapaul:
@eskabetxe @rmetzger can some of you take another look, please? :)

flink-connector-pinot/src/test/java/org/apache/flink/streaming/connectors/pinot/PinotTestBase.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        private static IndexingConfig getIndexingConfig() {
+            IndexingConfig indexingConfig = new IndexingConfig();
+            return indexingConfig;
+        }
+
+        private static TableCustomConfig getCustomConfig() {
+            TableCustomConfig customConfig = new TableCustomConfig(null);
+            ;
```
eskabetxe, line 220: abandoned ';'

eskabetxe (changes requested):
Sorry for late response..
LGTM, I left a minor comment, if you can change that I will merge once test are passing

mschroederi:
@eskabetxe Thanks for your feedback! Looking forward to seeing this merged.

eskabetxe:
@mschroederi thanks for your contribution
merged

PR #127, opened 2021-10-07
[BAHIR-284] add option to use token authentication for influxdb2

dquig:
See JIRA ticket: https://issues.apache.org/jira/browse/BAHIR-284


andre-casimiro:
Hi there, I'm interested in using this option but it's not available in the latest release `1.1.0`.
Any plans on creating a new release including that in the short term?

johncheng-crypto:
Without this option, I can't write to the influx sink. It throws com.influxdb.exceptions.UnauthorizedException

My env:
InfluxDB v2.3.0

andre-casimiro:
@johncheng-crypto I've made a copy of this project into mine to use the token authentication method, you could do the same while no release is out with it.

PR #128, opened 2021-10-07
[BAHIR-283] prevent each element from infludb2 sink buffer from being dropped

dquig:
See JIRA issue: https://issues.apache.org/jira/browse/BAHIR-283


eskabetxe:
@rmetzger could you check?

eskabetxe:
Thanks for your contribution @dquig

daniel-packard:
Hi @eskabetxe - 

Is there a release schedule for the next version of this sink?

currently using this version, which doesn't have the fix:
```
<dependency>
	<groupId>org.apache.bahir</groupId>
	<artifactId>flink-connector-influxdb2_2.12</artifactId>
	<version>1.2-SNAPSHOT</version>
	<scope>compile</scope>
</dependency>
```

eskabetxe:
Hi @daniel-packard,
I update the Snapshots, the fix should be available.

We are finishing the migration to Flink 1.17 when finished we should have a new release.

PR #130, opened 2021-11-09
[BAHIR-285] flink-connector-redis support SREM command.

wyxpku:
Support redis SREM command.


.github/workflows/maven-ci.yml
```diff
@@ -45,4 +45,4 @@ jobs:
       run: ./dev/change-scala-version.sh ${{ matrix.scala-version }}
       shell: bash
     - name: Build with flink ${{ matrix.flink-version }}
-      run: mvn -B clean verify -Dscala-${{ matrix.scala-version }} -Dflink.version=${{ matrix.flink-version }}
```
eskabetxe, line 48: why?

flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/RedisSinkITCase.java
```diff
@@ -77,14 +77,20 @@ public void testRedisListDataType() throws Exception {
...
             new RedisCommandMapper(RedisCommand.SADD));
 
-        source.addSink(redisSink);
-        env.execute("Test Redis Set Data Type");
+        source.addSink(redisSaddSink);
+        env.execute("Test SADD");
 
         assertEquals(NUM_ELEMENTS, jedis.scard(REDIS_KEY));
 
+        RedisSink<Tuple2<String, String>> redisSremSink = new RedisSink<>(jedisPoolConfig,
```
eskabetxe, line 88: could we add a new test only to remove?

eskabetxe (changes requested):
Could you separate in two commits one for each command added?

flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/RedisSinkITCase.java
wyxpku: > Could you separate in two commits one for each command added?

Sorry, It's a mistake, I'm new to github. 😅
I will close this PR and create a new one in future.

Thanks for review.

PR #139, opened 2021-12-15
[BAHIR-294] Bump log4j2 version to 2.17.0

deadwind4:
to fix CVE-2021-44228 and CVE-2021-45105
Although there is no place to use, we can prevent it in advance.


eskabetxe:
thanks @deadwind4  for your contribution

PR #142, opened 2022-01-04
Remove unnecessary jar packages commons-io 2.4

AsinRay:
flink-connector-redis using embedded-redis (version 0.6) for test. Although the scope is specified as test in the pom file, embedded-redis depends on common-io (version 2.4). Due to the transitivity of the dependency, the flink-connector-redis package depends on the unused common-io (version 2.4).
The common-io jar package needs to be excluded from the pom file.

[issue](https://issues.apache.org/jira/browse/BAHIR-298)

### DIFF
--- a/flink-connector-redis/pom.xml
+++ b/flink-connector-redis/pom.xml
@@ -55,12 +55,26 @@ under the License.
             <version>${jedis.version}</version>
         </dependency>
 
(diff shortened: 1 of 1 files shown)


AsinRay:
Remove unnecessary jar packages commons-io

PR #144, opened 2022-02-10
[BAHIR-299] Replace _2.11 with _${scala.binary.version}

1996fanrui:
https://issues.apache.org/jira/browse/BAHIR-299

When use scala-2.12, maven artifactId should result in _2.12, but a lot of bahir-flink code is hardcoded to _2.11. Using _${scala.binary.version} is better.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -20,13 +20,13 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.apache.bahir</groupId>
-        <artifactId>bahir-flink-parent_2.11</artifactId>
+        <artifactId>bahir-flink-parent</artifactId>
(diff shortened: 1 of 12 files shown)


1996fanrui:
Hi @eskabetxe , could you help to review this PR in your free time? Thanks a lot.

eskabetxe:
Hi @1996fanrui, thanks for your contribution..
We have a script to change the scala version (https://github.com/apache/bahir-flink/blob/master/dev/change-scala-version.sh)
Meanwhile, with the changes in flink 1.15 with "scala free" and the remove of all related scala versions from dependencies, I will prefer to wait and remove all the scala.version from dependencies too

@rmetzger , @lresende

1996fanrui:
Hi @eskabetxe Thanks a lot for your clarification. I will close this PR.

PR #155, opened 2022-08-25
[BAHIR-315] Redis connector: add Support for SSL connections for cluster

netsi:
* updated jedis to 4.3.0
* added support for SSL cluster connections

### DIFF
--- a/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
+++ b/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
@@ -47,14 +47,7 @@ public class RedisClusterContainer implements RedisCommandsContainer, Closeable 
     }
 
     @Override
-    public void open() throws Exception {
-
-        // echo() tries to open a connection and echos back the
-        // message passed as argument. Here we use it to monitor
-        // if we can communicate with the cluster.
-
-        jedisCluster.echo("Test");
-    }
+    public void open() throws Exception {}
 
     @Override
     public void hset(final String key, final String hashField, final String value, final Integer ttl) {
--- a/flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/RedisSinkTest.java
+++ b/flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/RedisSinkTest.java
@@ -27,7 +27,7 @@ import org.apache.flink.streaming.connectors.redis.common.mapper.RedisCommandDes
 import org.apache.flink.streaming.connectors.redis.common.mapper.RedisMapper;
 import org.apache.flink.util.TestLogger;
 import org.junit.Test;
-import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.exceptions.JedisException;
 
 import java.net.InetSocketAddress;
 import java.util.HashSet;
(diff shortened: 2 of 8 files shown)


amyxzheng:
hi @netsi how does the review process work for the changes in Bahir? We're trying out this Redis connector and hope to enable SSL

netsi:
@amyxzheng I can't answer this question, it is my first contribution attempt to this project. But there is some info in the README.md https://github.com/apache/bahir-flink#contributing-a-flink-connector . We're already using the version with the SSL connection and we didn't face any issues sofar.

@eskabetxe I've splitted the change into two commits, I'm not sure if it has to have 2 JIRA issues though. I also bumped the dependency to the latest one. I haven't tested it with the latest version though.
One thing that I didn't like when updating the dependency is that the `open()` function stopped functioning because the cluster connection doesn't support the `echo` function and I couldn't find an similar function that we could use, I've tried without success to get all the shards and execute the echo command there. Maybe someone can have a look at it.

flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/RedisSinkTest.java
```diff
@@ -78,7 +79,25 @@ public void testRedisClusterDownBehavior() throws Exception {
             .setMaxTotal(1)
             .setMinIdle(1).build();
 
-        testDownBehavior(wrongJedisClusterConfig);
+        RedisSink<Tuple2<String, String>> redisSink = new RedisSink<>(wrongJedisClusterConfig,
```
eskabetxe, line 82: why change this?
its the same code as testDownBehavior no?

flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
```diff
@@ -47,14 +47,7 @@ public RedisClusterContainer(JedisCluster jedisCluster) {
...
     @Override
-    public void open() throws Exception {
-
-        // echo() tries to open a connection and echos back the
-        // message passed as argument. Here we use it to monitor
-        // if we can communicate with the cluster.
-
-        jedisCluster.echo("Test");
-    }
+    public void open() throws Exception {}
```
eskabetxe, line 50: why the "jedisCluster.echo("Test")" is removed?

this could create any colateral? 
this is more an optimization than a need from the bump version no?

flink-connector-redis/src/test/java/org/apache/flink/streaming/connectors/redis/RedisSinkTest.java
netsi: The exception instance type has changed in the newer version. Previously the cluster and non cluster mode had the same type `JedisConnectionException` and now it is `JedisClusterOperationException`.

https://github.com/redis/jedis/blob/master/docs/3to4.md

```
All JedisCluster constructors now throw a JedisClusterOperationException if unable to connect to any of the provided HostAndPort(s). 
```

eskabetxe: I think we can change the exception from testDownBehavior to JedisException with our a problem..
or put the 2 exceptions on condition..

netsi: Good point. I've changed the condition to check for `JedisException`

eskabetxe:
@netsi thanks for your contribution, once tests pass I will merge this..

do you have an apache id, so I can assign the jira to you?

netsi:
@eskabetxe I don't have an apache id.  Not even sure how to get one :D

eskabetxe:
@netsi don't worry.. thanks for your contribution..
merged

amyxzheng:
hi @eskabetxe when will this change be released?

eskabetxe:
hi @amyxzheng currently is not planned any release soon..
we are working to update flink version to 1.15.2, but we have some test problems.. 

I could comment on create a new release with 1.14.6 but there are few changes since the last release

PR #156, opened 2022-10-06
Bump redis version to 4.2.9

srmrz:
Trying to expedite this change

https://github.com/apache/bahir-flink/pull/155

Splitting #155 into two separate changes.

cc @eskabetxe

### DIFF
--- a/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
+++ b/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
@@ -46,6 +46,7 @@ public class RedisClusterContainer implements RedisCommandsContainer, Closeable 
         this.jedisCluster = jedisCluster;
     }
 
+    @Override
     public void open() {}
 
     @Override
--- a/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
+++ b/flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
@@ -46,15 +46,7 @@ public class RedisClusterContainer implements RedisCommandsContainer, Closeable 
         this.jedisCluster = jedisCluster;
     }
 
-    @Override
-    public void open() throws Exception {
-
-        // echo() tries to open a connection and echos back the
-        // message passed as argument. Here we use it to monitor
-        // if we can communicate with the cluster.
-
-        jedisCluster.echo("Test");
-    }
+    public void open() {}
 
     @Override
     public void hset(final String key, final String hashField, final String value, final Integer ttl) {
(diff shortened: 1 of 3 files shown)


flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
```diff
@@ -46,15 +46,7 @@ public RedisClusterContainer(JedisCluster jedisCluster) {
...
-    @Override
-    public void open() throws Exception {
-
-        // echo() tries to open a connection and echos back the
-        // message passed as argument. Here we use it to monitor
-        // if we can communicate with the cluster.
-
-        jedisCluster.echo("Test");
-    }
+    public void open() {}
```
eskabetxe, line 49: Could you leave de override?

the "communication test" is not needed anymore?

eskabetxe (changes requested):

flink-connector-redis/src/main/java/org/apache/flink/streaming/connectors/redis/common/container/RedisClusterContainer.java
srmrz: re:override, sure, let me do that!

I'll take a look at the "communication test"

srmrz: @eskabetxe, added override here [94e2078](https://github.com/apache/bahir-flink/pull/156/commits/94e2078feec7b98d22a083b23c9a826a836c6d80)

Mind pointing me at the communication test?

eskabetxe: @srmrz "jedisCluster.echo("Test");"
I don't know if is needed.. but for the comment its testing the communication with cluster..
I don't write this code, I'm trying to understand if removing that code have some side effect..

srmrz: @eskabetxe I see! 

> I don't know if is needed

My take is that it isn't needed - the constructor takes a `JedisCluster`. 

As part of the initialization of `JedisCluster` , it contacts the cluster:

https://github.com/redis/jedis/blob/379bbed0ee363bfc9f31e4591f5d46c28976cff5/src/main/java/redis/clients/jedis/UnifiedJedis.java#L105-L107

https://github.com/redis/jedis/blob/379bbed0ee363bfc9f31e4591f5d46c28976cff5/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java#L27

https://github.com/redis/jedis/blob/379bbed0ee363bfc9f31e4591f5d46c28976cff5/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java#L36-L49

That runs commands in the cluster. So if there are problems contacting the cluster, we'll know before we even construct a `RedisClusterContainer`

srmrz: Checking in on this. Let me know if there are other questions

PR #159, opened 2022-10-21
[BAHIR-315] Move SSL config to FlinkJedisConfigBase

srmrz:
When trying to use the recently added support for SSL (#155), we realized that we needed support for non-cluster mode.

Here we add support for SSL connections in non-cluster configurations.

Two things:
1. JedisSentinelPool currently doesn't support SSL connections. So I'm opting so I'm forcing ssl=false for that configuration.
2. Renamed `ssl` to `useSsl`. But I'm very happy to change that back to `ssl` if there's pushback.


srmrz:
Let me know if this can be considered part of https://issues.apache.org/jira/browse/BAHIR-315 or if should create a separate ticket.

eskabetxe:
Hi @srmrz ,
Sorry for the delay..
Thanks for your contribution..
Merged

PR #166, opened 2023-05-30
BAHIR-324  JobManager created a kuduReader that was not closed, resulting in an increase in server handles

collabH:
# ISSUE Location
[ISSUE BAHIR-324](https://issues.apache.org/jira/projects/BAHIR/issues/BAHIR-324?filter=allopenissues&orderby=priority+DESC%2C+updated+DESC)
# Problem
When JobManager starts, the ExecutionJobVertex constructor calls the AbstractKuduInputFormat#createInputSplits method, Causes the KuduReader object to create the KuduClient and KuduSession objects, but ExecutionJobVertex does not call the Kudureader #close() method with the display, causing the Kudu-related handle to remain open after the Flink task is closed. As a result, the number of server handles increases with the number of Flink tasks. As a result, the entire server becomes unavailable

### DIFF
--- a/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/format/AbstractKuduInputFormat.java
+++ b/flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/format/AbstractKuduInputFormat.java
@@ -104,6 +104,13 @@ public abstract class AbstractKuduInputFormat<T> extends RichInputFormat<T, Kudu
         }
     }
 
+    private void closeKuduReader() throws IOException {
+        if (kuduReader != null) {
+            kuduReader.close();
+            kuduReader = null;
+        }
+    }
+
     @Override
     public void close() throws IOException {
         if (resultIterator != null) {
@@ -113,10 +120,7 @@ public abstract class AbstractKuduInputFormat<T> extends RichInputFormat<T, Kudu
                 e.printStackTrace();
             }
         }
-        if (kuduReader != null) {
-            kuduReader.close();
-            kuduReader = null;
-        }
+        closeKuduReader();
     }
 
     @Override
@@ -131,8 +135,12 @@ public abstract class AbstractKuduInputFormat<T> extends RichInputFormat<T, Kudu
 
     @Override
     public KuduInputSplit[] createInputSplits(int minNumSplits) throws IOException {
-        startKuduReader();
-        return kuduReader.createInputSplits(minNumSplits);
+        try {
+            startKuduReader();
+            return kuduReader.createInputSplits(minNumSplits);
+        } finally {
+            closeKuduReader();
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


collabH:
Hi @eskabetxe , could you check?

eskabetxe:
Hi @collabH,

could we have a test for this?

collabH:
> Hi @collabH,
> 
> could we have a test for this?
https://github.com/apache/bahir-flink/blob/176242c9d10d8a67a0f773dc61235b96b26e3c44/flink-connector-kudu/src/test/java/org/apache/flink/connectors/kudu/format/KuduRowDataInputFormatTest.java#L63

Please see KuduRowDataInputFormatTest#testInputFormatWithProjection method, this part has been covered

eskabetxe:
> > Hi @collabH,
> > could we have a test for this?
> > https://github.com/apache/bahir-flink/blob/176242c9d10d8a67a0f773dc61235b96b26e3c44/flink-connector-kudu/src/test/java/org/apache/flink/connectors/kudu/format/KuduRowDataInputFormatTest.java#L63
> 
> Please see KuduRowDataInputFormatTest#testInputFormatWithProjection method, this part has been covered

but that test dont check that connections are closed no? that is the problem you are trying to solve with this PR..

collabH:
> > > Hi @collabH,
> > > could we have a test for this?
> > > https://github.com/apache/bahir-flink/blob/176242c9d10d8a67a0f773dc61235b96b26e3c44/flink-connector-kudu/src/test/java/org/apache/flink/connectors/kudu/format/KuduRowDataInputFormatTest.java#L63
> > 
> > 
> > Please see KuduRowDataInputFormatTest#testInputFormatWithProjection method, this part has been covered
> 
> but that test dont check that connections are closed no? that is the problem you are trying to solve with this PR..

It is not feasible to use test code to verify this problem at present, but we ran the previous Flink task according to the repaired code in our own production environment. After observing that the corresponding Kudu handle was also closed after the Flink task was run, we have fixed this problem according to the reasons previously located.

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/format/AbstractKuduInputFormat.java
```diff
@@ -113,10 +120,7 @@ public void close() throws IOException {
                 e.printStackTrace();
```
eskabetxe, line 120: O know this is unrelated, but can we change this to a log?

eskabetxe (changes requested):

flink-connector-kudu/src/main/java/org/apache/flink/connectors/kudu/format/AbstractKuduInputFormat.java
collabH: sure.

eskabetxe:
Merged

